PR #21, opened 2016-01-22
two commands, per your request.

horstm22:


rjbs:
Thanks, applied!  It needed some tweaking to work, you may or may not be interested in reviewing 4afbd57c72767666b0e34bab5bd06ff70b7166fe and subsequent commits.

PR #22, opened 2016-08-10
Don't send an sms on reminders unless asked for.

wolfsage:


rjbs:
Wow, good find! But can you instead make page the default and no-page the extra?  I <3 the pages.

wolfsage:
Sure thing, will tackle later

wolfsage:
Yup! Tested a few scenarios, should be good!

rjbs:
merged, applied, live

Issue #15, opened 2014-01-21
give a little grace before nagging after a timer is stopped


rjbs:
Something like: don't nag if we saw that a timer *was* running fewer than 10 minutes ago.

PR #23, opened 2016-08-18
Don't nag a user about a timer not running if they recently had one.

wolfsage:
The default grace period is 15 minutes, but can be changed by
setting nag_grace_period in config.

A value of less than 5 minutes is useless since timers are checked
every 5 minutes anyway.

This addresses #15


rjbs:
Thanks, applied!

PR #25, opened 2016-11-29
Allow assigning tasks to a project

wolfsage:
++ Some task #project
    ++ Other task #"Another project"

Right now this assumes all users will have access to the same
list of projects, but we could cache projects per-user.

Another complication is that if #something appears in the task name it will be yanked out as the project.

I wonder if we should have some delimiter before metadata like this that will never be allowed to appear as part of the task.

Then we could have other keywords as well to set things like packages and estimates. Something like:

    ++ Don't allow such and such -- #project @package

### DIFF
--- a/synergy
+++ b/synergy
@@ -13,6 +13,7 @@ use POE::Component::Server::SimpleHTTP;
 use Time::Duration::Parse;
 use Time::Duration;
 use YAML::XS;
+use Data::Dumper;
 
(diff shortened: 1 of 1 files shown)


wolfsage:
Closing this for now; this doesn't allow setting things in sub folders if they share names with other sub folders of different parents, and really it'd be nice to set a whole host of things besides just the project....

PR #26, opened 2016-12-26
saved state!

rjbs:
Add a saved state file and use it to persist timer and reminders.

### DIFF
--- a/lib/Synergy/Timer.pm
+++ b/lib/Synergy/Timer.pm
@@ -16,6 +16,10 @@ has chilltill => (
   clearer   => 'clear_chilltill',
 );
 
+after clear_chilltill => sub {
+  POE::Kernel->yield('save_state');
+};
+
 sub chilling {
   my ($self) = @_;
   return 1 if $self->chill_until_active;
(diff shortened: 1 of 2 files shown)


wolfsage:
Seems legit.

yield() to save state could mean we send out reminders/timers then crash before saving when handling some other event maybe....

rjbs:
I considered that, and here's my thinking, more or less:

* if we block on saving, then send reminders, we could fail to send but think we did
* if we block on sending, then save, that's too blocking, and we could send twice

So the best solution is probably to do something like `{ send_reminder }->and_then({ save_state })`

This gets a bit more complicated, because if send_reminder removes the reminder from the reminders list, an intermediate state-save (like for a chill) will write the removed reminder.  So we actually want send_reminder to do the removal of the item, but that means that check-reminders might run twice and find the same reminder in the queue, not yet removed by send_reminder, which is still waiting.  So we probably want the yield to send_reminder-then-save-state to mark the job as taken in such a way that it won't be saved to state, but will prevent second-enqueueing of the send-reminder.

That's not too hard, I reckon.  We give a timer a "taken" property that isn't saved.  Sound about right for take 2?

wolfsage:
Good points.

yes, I think that sounds right.

PR #28, opened 2016-12-27
per-user time zones

rjbs:
Give each user a TZ.  Try to use it when showing them a time.

### DIFF
--- a/synergy
+++ b/synergy
@@ -661,7 +661,7 @@ sub SAID_good {
   }
 
   if ($end_of_day && (my $sy_timer = $user->timer)) {
-    my $time = $self->_parse_time_hunk('until tomorrow');
+    my $time = $self->_parse_time_hunk('until tomorrow', $user);
     $sy_timer->chilltill($time);
   }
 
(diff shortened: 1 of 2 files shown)


synergy
```diff
@@ -814,18 +814,19 @@ sub SAID_shows {
...
-  },
-);
+sub date_parser_for {
+  my ($self, $user) = @_;
+
+  my $tz = $user ? $user->time_zone : 'America/New_York';
+
+  DateTime::Format::Natural->new(
+    prefer_future => 1,
+    time_zone     => 'America/New_York',
```
wolfsage: Need to use $tz here

synergy
```diff
@@ -876,29 +878,40 @@ sub SAID_chill {
...
   POE::Kernel->yield('save_state');
-  $self->reply("Okay, no more nagging until " . localtime $time, $arg);
+  $self->reply("Okay, no more nagging until $when", $arg);
 }
 
 sub SAID_mute {
   my ($self, $arg) = @_;
   return -1 unless $arg->{what} =~ s/\A\s*(alerts|all|\*)\s+//;
 
+  my $user = $self->resolve_name($arg->{who});
```
wolfsage: Always guaranteed to get a user back?

synergy
```diff
@@ -936,6 +949,8 @@ sub SAID_remind {
 
   $who = $self->resolve_name($who, $arg->{who});
 
+  my $speaker = $self->resolve_name($arg->{who});
```
wolfsage: silly bug but with this usage, resolve_name will return nothing if $arg->{who} is someone named 'myself', or 'my', etc...

synergy
rjbs: Yes.  It's stupid: we've already resolved the user when setting up the `$arg` that's passed in, but we pass the username, not object.  Historical reasons.  I plan to fix this.

synergy
rjbs: Can't happen unless we set up such users, which I promise not to do. :-)

rjbs:
Thanks, fixed and merged.

PR #29, opened 2017-01-05
Add a cpanfile and stupidly simple example config

perigrin:
So the two lowest hanging fruit were a list of dependencies and the undocumented config file. This should fix both of those.


rjbs:
Thanks, applied!

PR #30, opened 2017-01-18
Allow abort/reset to work without timer arg, and be more helpful

wolfsage:


rjbs:
I am hesitant to allow bare "abort," but since I don't use it, I won't be the one inconvenienced later when I forbid it!  :D  Applied.

PR #31, opened 2017-01-18
Be more helfup when program is run with no/bad config file

wolfsage:


rjbs:
Thanks, applied and subsequently slightly tweaked in the textual area.

PR #32, opened 2017-04-11
expand help for more commands

horstm22:


rjbs:
Thanks, applied!

I undocumented "addchecklist" for now, because I'm not sure we really want to use it.  Do you think we do?

horstm22:
I assumed it was there because you used it! And if you use it and find it helpful, others would, too. Feel free to leave out if it was a lark.

PR #34, opened 2017-06-08
Add 'unestimated' command - used just like 'damage'

wolfsage:
But this will return a list of all of URLs to all of your
unestimated tasks in a private response.


rjbs:
merged, thanks

PR #35, opened 2017-06-08
Detect unestimated better

wolfsage:
This builds on https://github.com/rjbs/Synergy/pull/34 but doesn't necessarily require it.

It could be adapted to master if #34 isn't merged - the important bit is detecting unestimated tasks properly


rjbs:
merged, thanks;

PR #36, opened 2017-06-09
Set max_connections_to_host to 5 for conns to liquid planner

wolfsage:
Otherwise each lp request will queue up behind the currently
firing request. This number can be increased to whatever limit
LiquidPlanner may have (or set to 0 to be unlimited).


rjbs:
merged, thanks

PR #38, opened 2017-06-20
Let us know when we fail to send a page for some reason

wolfsage:
Was only able to test failure modes; please test successes!


rjbs:
merged this a while ago

PR #45, opened 2017-07-20
Allow "spent time on <task_id>" to log time against task

jwshea:
Also support updating a task where the first numeric match is a task_id


rjbs:
Thanks!  Sorry I missed this when it came in.  It's now live.

PR #47, opened 2017-08-17
Add task name in response to 'spent' command

mmcclimon:


rjbs:
Thanks, @mmcclimon! 💙

PR #48, opened 2017-08-17
Add link to LP tickets when mentioned in passing

mmcclimon:
This is like what bort does, only for LiquidPlanner IDs.

This is a little bit different that what synergy normally does, since she'll respond like this even if you don't mention her.


rjbs:
Thanks, @mmcclimon! 💙

PR #49, opened 2017-10-17
Add highfive support to synergy. Requires slack integration

wolfsage:
Set up slack_highfive/token if you want incoming slack webhooks to work
Set up slack_highfive/url if you just want irc->slack highfives to work


rjbs:
This was merged in v1 and lost in v2.

PR #50, opened 2017-11-17
You can now create tasks in specific projects

wolfsage:
These projects must be specified in the config file under 'projects':

projects:
  short_name_no_spaces: project_id
  other: id2

Where the ids are the numerical ids from liquidplanner.

This adds one command 'projects' which will dispaly all known projects in
private message (and a link to them in liquid planner).

To use the functionality:

  ++ some task #project_short_name
  >> someone: fix foo #project_short_name
  task for so and so: whatever #project_short_name

Note that if # is seen in the name, we assume everything after it is
a project to prevent mishaps. In this case, the project *must exist* or
we reject the creation.

### DIFF
--- a/synergy
+++ b/synergy
@@ -94,6 +94,29 @@ has users => (
   writer  => '_set_users',
 );
 
+has projects => (
+  isa => 'HashRef',
+  traits => [ 'Hash' ],
+  handles => {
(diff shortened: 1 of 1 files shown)


rjbs:
This was merged, more or less.

PR #51, opened 2017-11-21
spent x on #taskid (>) should start a timer

wolfsage:


rjbs:
This PR is against v1, but this feature is in v2.

wolfsage:
Should we port it?

PR #52, opened 2017-11-21
Good bye

wolfsage:
Maybe not so useful


rjbs:
Warp core ejection does not exist in v2.

PR #54, opened 2018-02-19
When stopping timer, always say task name

mmcclimon:
Before, this was only said if you marked the task done. That seems
silly.


mmcclimon:
I have done this myself in 7b61367  using my all-powerful push rights.

PR #55, opened 2018-04-15
Steal bort's Gitlab functionality

mmcclimon:
This steal's bort's mention mr-or-commit in passing. Project shortcuts are gotten from our config repo. (Not committing this directly to master because we probably don't want it until we turn bort's off!)

### DIFF
--- a/lib/Synergy/Reactor/GitLab.pm
+++ b/lib/Synergy/Reactor/GitLab.pm
@@ -25,6 +25,13 @@ has api_uri => (
   required => 1,
 );
 
+has url_base => (
+  is      => 'ro',
+  isa     => 'Str',
+  lazy    => 1,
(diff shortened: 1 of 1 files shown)


wolfsage:
Seems good.

We don't mark an event handled though if we encounter an error trying to handle it? Should we?

mmcclimon:
> We don't mark an event handled though if we encounter an error trying to handle it? Should we?

I don't think so. All that `$event->mark_handled` does (at the moment)  is trigger the 'does not compute' response. In a public channel, we don't want that response, since it's just a mention-in-passing. If it's targeted or in a DM and we don't handle it, we still want "does not compute" so that it's not totally silent.

mmcclimon:
@rjbs Amended and updated!

rjbs:
Applied!  Redundant until Bort is silenced.

PR #57, opened 2018-04-26
Hack around bad IO::Async::SSL behaviour so we can work with SNI hosts

wolfsage:
IO::Async::SSL has a bug (I think) preventing it from working with hosts
that require SNI. This fixes that by forcing the hostname to the requested
hostname.


rjbs:
This was merged as 68c8d0d.

PR #58, opened 2018-06-05
Queue messages if we're in the process of reconnecting

wolfsage:
Then flush them once we've reconnected.

That way, we don't lose any data if we happen to drop our
connection after we've made some changes but before we've
responded to a request.


mmcclimon:
LGTM.

PR #62, opened 2018-06-13
Add 'upgrade' reactor which will git pull and then exit

wolfsage:


rjbs:
* would rather fetch and reset than pull
* if there are dirty files, refuse
* this makes me a little uncomfortable
* but what I've been doing in production is worse

PR #64, opened 2018-06-29
Allow true async requests in Synegy and Synergy::LPC

wolfsage:
We do this using futures. This means Synergy::LPC isn't tied
to IO::Async (or any event framework).

Right now async is just used in damage report to speed up
its operations since it makes a number of calls (and one of
them is very slow).


wolfsage:
BEHOLD! THE FUTURES!

rjbs:
Needs a rebase. :(

wolfsage:
Rebased :(

rjbs:
LiquidPlanner is dead.

PR #65, opened 2018-07-01
Remove LP-specific stuff from Synergy::User

mmcclimon:
This rightfully belongs in the LP reactor! This actually uses the
preferences now, and I think it even works.

@rjbs, would like a second set of eyes on this, since if I broke it I break all of LP!


rjbs:
I'll talk to you about this tomorrow.  I have questions and/or misgivings.

PR #71, opened 2018-08-09
Prometheus support

robn:
You could merge #69 first, or just do it all here!


rjbs:
This live for a few days now!

PR #75, opened 2018-08-13
HTTPS server support

robn:
Don't much care for explicitly threading all the config options through, and it probably wants more config options for non-trivial uses, but I just did the dumbest thing first.


lib/Synergy/Hub.pm
```diff
@@ -104,14 +105,26 @@ has server_port => (
...
   is => 'ro',
   isa => 'Synergy::HTTPServer',
   lazy => 1,
   default => sub ($self) {
     my $s = Synergy::HTTPServer->new({
-      name        => '_http_server',
-      server_port => $self->server_port,
+      name          => '_http_server',
+      server_port   => $self->server_port,
+      defined_kv(tls_cert_file => $self->tls_cert_file),
```
mmcclimon: (Leaving aside whether we want to do the HTTPS stuff in synergy or elsewhere...) I'd probably give these a default and then just always pass them. Server.pm will ignore a blank one anyway.

lib/Synergy/Hub.pm
```diff
@@ -295,6 +308,8 @@ sub synergize {
     user_directory  => $directory,
     ($config->{time_zone_names} ? (time_zone_names  => $config->{time_zone_names})  : ()),
     ($config->{server_port}     ? (server_port      => $config->{server_port})      : ()),
+    defined_kv(tls_cert_file => $config->{tls_cert_file}),
```
mmcclimon: I like defined_kv as much as the next guy (unless the next guy is you, in which case you probably like it more), but we should use it on all or none, not half!

lib/Synergy/HTTPServer.pm
```diff
@@ -85,6 +97,17 @@ sub start ($self) {
     on_listen_error => sub { die "Cannot listen - $_[-1]\n" },
   );
 
+  if ($self->tls_cert_file) {
```
mmcclimon: Should probably also check for tls_key_file here, just to be sure.

lib/Synergy/Hub.pm
robn: Fair. Dumb style question then, would you do it in the same commit? I hate it when a commit is the stated changes plus one or two other unrelated things, but maybe this is fine.

I'm just gonna do it, so don't think very hard :)

rjbs:
The `""` for unset values troubles me, especially since we're using `defined_kv` here and there.  We should stick to "it's unset" or "it's set to a valid filename".

Other than that… looks good!

robn:
Seems I fixed all this but didn't push. Now I have!

PR #77, opened 2018-08-24
Be case-sensitive about responding to “RFC”

chris-morgan:
I expect “rfc\d+” to be entirely false positives.

This is probably most notable in the MIME type “message/rfc822” which
crops up fairly often in the email space. If this approach of requiring
capital “RFC” is considered too radical, an alternative would be to
specifically not respond to “rfc822” (case sensitive). But that’s more
work to implement.


wolfsage:
I think it'd be better to say "(^|\s)rfc" or something and keep it case insensitive.

chris-morgan:
I’m going to leave any finer decisions to others.

PR #78, opened 2018-08-31
HTTPServer: authentication support

robn:
Adds optional auth to HTTP endpoints. Set in the config file on a per-path basis:

```
"http_auth" : {
    "/somepath" : [ "admin", "secret" ],
    "otherpath" : [ "jefferson", "ford" ]
}
```

Paths not listed are unauthenticated as you'd expect.

Maybe it'd be better to wrap the app at registration time, but this makes it a little easier to reconfigure on the fly later (only have to modify the `http_auth` hash, not rewrap everything).

Still not thrilled about threading the config vars through this way but I'd rather solve that some other time.


lib/Synergy/HTTPServer.pm
```diff
@@ -71,7 +82,26 @@ has http_server => (
           return $req->new_response(404)->finalize;
         }
 
-        return $self->app_for_path($req->path_info)->($req);
+        my $app = $self->app_for_path($req->path_info);
+
+        if ($self->path_requires_auth($req->path_info)) {
+          my $creds = $self->auth_credentials_for_path($req->path_info);
+          my ($authuser, $authpass) = $creds->@*;
```
wolfsage: Yeah would much prefer key/value pair for the creds

t/httpserver.t
```diff
@@ -48,5 +54,30 @@ my $port = $synergy->server_port;
...
+{
+  my ($res) = $http->do_request(
+    uri => "http://localhost:$port/auth",
+    headers => {
+      'Authorization' => "Basic ".MIME::Base64::encode_base64("admin:badpass", "")
+    }
+  )->get;
+  is($res->code, 401, 'http request to auth endpoint with wrong creds returns 401');
+}
+
```
wolfsage: Wooh tests!

wolfsage:
One thing I don't love is it separates the registration of paths from the specification of which paths require auth - which seems like it could lead to typoing the auth name or something.

It's fine for now I think, but we gotta be careful.

wolfsage:
+1

robn:
> One thing I don't love is it separates the registration of paths from the specification of which paths require auth - which seems like it could lead to typoing the auth name or something.
> 
> It's fine for now I think, but we gotta be careful.

Mm. I think at minimum, I should have it warn/die if you specified auth for any paths that don't exist. I will do that.

Better in the future would probably be to have a common config for http path+auth in individual reactors & channels. Maybe just some role so its all automatic and checked and stuff. I have not thought about this.

wolfsage:
> Mm. I think at minimum, I should have it warn/die if you specified auth for any paths that don't exist. I will do that.

Yeah, I was thinking about that. You'd have to do something delayed I think since registering app locations can happen sort of any time but auth happens at config time?

t/httpserver.t
robn: I am being _very_ well behaved in this repo 😁

lib/Synergy/HTTPServer.pm
robn: Done.

robn:
> > Mm. I think at minimum, I should have it warn/die if you specified auth for any paths that don't exist. I will do that.
> 
> Yeah, I was thinking about that. You'd have to do something delayed I think since registering app locations can happen sort of any time but auth happens at config time?

Mm, good tip. I will consider further this evening from the couch.

robn:
Have considered, and I don't think there's an obvious place to do the checks as its currently structured. As you say, registering http paths can happen any time, so there's no way for us to know if an unused auth config will apply in the future.

I'm fiddling with a HTTP endpoint role tonight. If it pans out, it'll be much better than this :)

PR #80, opened 2018-10-05
Convert LP timers to objects. Use *real total time* when committing work

wolfsage:
A timer has a total_time and a running_time.

If you stop a timer then later resume it, the total_time is the
total previous time that the timer has accumulated, *not including*
the current running_time of the timer.

So to get the real running time of a timer, we need a new method:

  real_total_time()

which combines these two.

I've also added convenience methods to get at friendly readings
of the 3 different times:

  running_time_duration
  total_time_duration
  real_total_time_duration

This gives you something like "1h5m".


rjbs:
I gave it a quick look, it seems good to me.

PR #84, opened 2018-10-29
Add a HighFive reactor for slash commands and in-chat messages

wolfsage:


lib/Synergy/Reactor/HighFive.pm
```diff
@@ -0,0 +1,197 @@
...
+    }
+  );
+
+  return $res;
+};
+
+sub listener_specs {
+  return {
+    name      => 'highfive',
+    method    => 'high_five',
```
mmcclimon: `highfive` vs `high_five` seems silly and arbitrary, though I totally understand why you did it.

lib/Synergy/Reactor/HighFive.pm
```diff
@@ -0,0 +1,197 @@
...
+  };
+}
+
+sub start ($self) {}
+
+sub high_five ($self, $event) {
+  $event->mark_handled;
+
+  unless ($event->from_user) {
+    $event->reply("Sorry, I don't know who you are.");
```
mmcclimon: Needs a `return`.

lib/Synergy/Reactor/HighFive.pm
```diff
@@ -0,0 +1,197 @@
...
+  return $res;
+};
+
+sub listener_specs {
+  return {
+    name      => 'highfive',
+    method    => 'high_five',
+    predicate => sub ($, $e) {
+      $e->was_targeted &&
+      $e->text =~ /^highfive.*$/i;
```
mmcclimon: (nit) This `.*` is unnecessary

lib/Synergy/Reactor/HighFive.pm
```diff
@@ -0,0 +1,197 @@
...
+  my $success = $arg{success} || sub {};
+  my $failure = $arg{failure} || sub {};
+
+  # "highfive to rjbs for eating pie" or "highfive for @rjbs for baking pie"
+  $text =~ s/\Ahighfive\s*(to|for)\s+//i;
+
+  my ($target, $reason) = split /\s+/, $text, 2;
+
+  my $target_user = $self->hub->user_directory->user_named($target);
+  unless ($target_user) {
```
mmcclimon: This is totally fine, but I think current highfive behavior allows you to send a highfive to anything, and new behavior won't.

That is, you can say `/highfive @phl for being great` or `/highfive tests for existing`

lib/Synergy/Reactor/HighFive.pm
```diff
@@ -0,0 +1,197 @@
...
+}
+
+sub do_high_five ($self, %arg) {
+  my $from = $arg{from};
+  my $text = $arg{text};
+  my $chan = $arg{chan} || '';
+  my $success = $arg{success} || sub {};
+  my $failure = $arg{failure} || sub {};
+
+  # "highfive to rjbs for eating pie" or "highfive for @rjbs for baking pie"
```
mmcclimon: I feel like `@synergy highfive rjbs: eating pie` should work, without the "to" or "for"

lib/Synergy/Reactor/HighFive.pm
wolfsage: I meant to standardize. Can probably just do "highfive" and be done with it

lib/Synergy/Reactor/HighFive.pm
wolfsage: Right right thanks

lib/Synergy/Reactor/HighFive.pm
wolfsage: Oops

lib/Synergy/Reactor/HighFive.pm
wolfsage: Oh good call

lib/Synergy/Reactor/HighFive.pm
wolfsage: It does. This regex is just cleaning up some of the cases

wolfsage: Ah no, it doesn't. Blarg.

rjbs:
I think this LGTM, but!  Right now, we generate an SQLite db entry for each highfive.  I will give you the code.  This is used for the newsletter.  We need to keep that.  With that done, we'll cut over.

```perl
    my $dbh = DBI->connect(
      "dbi:SQLite:dbname=/home/rjbs/.synergy/synergy.sqlite",
      undef,
      undef,
    );

    $dbh->do(
      "INSERT INTO highfives
      (from_user, to_user, reason, highfived_at)
      VALUES
      (?, ?, ?, ?)",
      undef,
      $who,
      $target,
      $reason // "",
      time,
    );
```

PR #85, opened 2018-11-02
LP: add "assign" and "reassign" commands

robn:
"assign" adds an assignment for the named user, nothing more.

"reassign" is a bit more powerful. It replaces your assignment (if you had one) with the target, or if its assigned to one other person, replaces their assignment with the target (that can be you!)


rjbs:
LiquidPlanner is dead.

PR #87, opened 2018-11-28
UserDictionary: add pronoun preference

robn:


mmcclimon:
Applied with some tweaks, thanks!

PR #89, opened 2019-02-02
Respond to slack edits

mmcclimon:
This might actually be ok, though it needs a bit of polish. I have done basically no testing with non-slack channels yet. The first commit here is almost entirely whitespace.

### DIFF
--- a/lib/Synergy/External/Slack.pm
+++ b/lib/Synergy/External/Slack.pm
@@ -99,6 +99,20 @@ has _team_data => (
   writer => '_set_team_data',
 );
 
+has pending_frames => (
+  is => 'ro',
+  isa => 'HashRef',
+  lazy => 1,
+  default => sub { {} },
+);
+
+has pending_timeouts => (
+  is => 'ro',
+  isa => 'HashRef',
+  lazy => 1,
+  default => sub { {} },
+);
+
 sub connect ($self) {
   $self->connected(0);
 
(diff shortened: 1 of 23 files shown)


lib/Synergy/External/Slack.pm
```diff
@@ -115,17 +109,51 @@ sub connect ($self) {
             ->get;
 };
 
+my %pending_frames;
+my %pending_timeouts;
```
rjbs: These should be object, not global state.

lib/Synergy/Reactor/Upgrade.pm
```diff
@@ -124,14 +124,8 @@ sub handle_upgrade ($self, $event) {
...
 
-  my $timer = IO::Async::Timer::Countdown->new(
-    delay => 1,
-    on_expire => sub { kill 'INT', $$ }, # Why not just exit?
-  );
-
-  $self->hub->loop->add($timer);
-
-  $timer->start;
+  # Why not just exit?
```
wolfsage: Moving this comment here changes the intent of what it's asking. It goes from "Why are we using `kill 'INT', $$`?" to "Why are we delaying a second and then killing ourselves".

It might be worth stating outright why we delay (so our replies to the user have time to make it to slack) and then asking "why don't we just exit" or finally just changing that to exit.

Really, though, what we want is to be able to say "After this reply has successfully gone to slack it is safe to exit".

PR #90, opened 2019-02-04
Async http

mmcclimon:
This steals the async http bits from #64, then uses them to make the Gitlab reactor fully asynchronous. There are some other relatively minor changes to Synergy::Event and Synergy::Hub to make it more futurey.


mmcclimon:
I merged this!

PR #94, opened 2019-02-18
Yelling: yell louder, yell harder

mmcclimon:
This is a hilarious, terrible hack so that Synergy will ensure that all
plain text delivered to #yelling is in all caps.

### DIFF
--- a/lib/Synergy/Event.pm
+++ b/lib/Synergy/Event.pm
@@ -114,12 +114,16 @@ sub error_reply ($self, $text, $alts = {}) {
   return $self->reply($text, $alts, { was_error => 1 });
 }
 
+sub get_reply_prefix ($self) {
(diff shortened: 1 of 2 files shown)


mmcclimon:
jk, all this does is make the prefix uppercase.

mmcclimon:
Totes updated. No problem can't be solved with more metaprogramming.

robn:
Its funny, because this kind of makes me squeamish in the same way that futzing in the symbol table makes me squeamish. At the same time it doesn't, because this sort of shenanigans is exactly the reason that the MOP exists, and this is so much better anyway because you can choose to modify a single instance of a class rather than the entire world.

So yeah, I like this!

mmcclimon:
Closing this in favor of 6bf84766.

PR #96, opened 2019-02-28
WIP: Discord channel

robn:
Very early [Discord](https://discordapp.com/) channel, based on the Slack channel. So far handles group and direct channels and simple chatter. Lots more to do to reach feature parity.

screen shot 2019-02-28 at 9 57 04 pm

### DIFF
--- a/lib/Synergy/Channel/Discord.pm
+++ b/lib/Synergy/Channel/Discord.pm
@@ -57,7 +57,7 @@ sub handle_discord_event ($self, $name, $event) {
     $Logger->log("discord: ignoring message, we aren't ready yet");
     return;
   }
-  
+
   my $synergy_event = $self->synergy_event_from_discord_event($event);
   $self->hub->handle_event($synergy_event);
 }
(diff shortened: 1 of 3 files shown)


rjbs:
Nice.

1. I'm gonna assume this shouldn't be merged while marked WIP. ;-)
2. I have always been a little unsure about the "External" thing.  Let me know if you have a cooler idea.  Something related to making a standalone notifier, maybe.

mmcclimon:
Yeah, the External::Slack stuff is mostly a relic of what I started with and didn't get properly incorporated in the 3-day synergy rewriteathon. But on the other hand, it's a bit useful to keep all the low-level frame munging outside of the channel, maybe? Dunno.

robn:
Yeah, not for merge. I mean you could, and it wouldn't hurt anything, but its not super-useful as-is.

In my mind the channel/external split was that external was more-or-less a standalone client, with all the frame munging etc, as @mmcclimon says. They're a little bit entangled on the Slack side (`::start` sets the `on_frame` handler), but I think the idea is mostly ok.

There might be a line under where you can shift the Slack/Discord stuff, and then have a single higher-level "channel" that does work for both, but it would need a lot of thinking. I reckon Slack and Discord are close enough in function, at least how we use it, but dunno if anything else is. Maybe wait for the third system.

Things that would have helped, that we've totally had no reason to ever build and maybe never will, is tests that exercise the channel interface. There's still stuff I haven't implemented, that I'm having to find reactors that use it. Some stuff is only used by LP currently, which of course I dare not hook up to Discord.

rjbs:
Good news: this rebases just fine.

rjbs:
I have pushed the rebase here, plus a commit to make it run again.

rjbs:
This has long since been merged!

PR #97, opened 2019-03-16
Self-registration!

mmcclimon:
This converts the user directory into a set of tables in the sqlite database. It'll auto-upgrade from an existing user-directory file on the fly.

The only thing in the user table that probably shouldn't be is `lp_id`, but removing that was gonna be a pain because we also have virtual users who can't self-register. That's a problem for another day.

### DIFF
--- a/lib/Synergy/Channel/Pushover.pm
+++ b/lib/Synergy/Channel/Pushover.pm
@@ -29,7 +29,7 @@ sub http_post {
 }
 
 sub send_message_to_user ($self, $user, $text, $alts = {}) {
-  unless ($user->identities->{ $self->name }) {
+  unless ($user->has_identity_for($self->name)) {
     $Logger->log([
       "can't send message, no api key for %s",
       $user->username,
@@ -37,7 +37,7 @@ sub send_message_to_user ($self, $user, $text, $alts = {}) {
     return;
   }
 
-  my $where = $user->identities->{ $self->name };
+  my $where = $user->identity_for($self->name);
   my $who = $user->username;
 
   $Logger->log([ "sending pushover <$text> to $who" ]);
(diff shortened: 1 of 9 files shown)


lib/Synergy/Reactor/Register.pm
```diff
@@ -0,0 +1,73 @@
...
+
+  return $event->error_reply("Sorry, usernames must be all letters")
+    if $username =~ /[^A-Za-z]/i;
+
+  die "crazy case: event has no from address??" unless $event->from_address;
+
+  if ($directory->user_named($username)) {
+    my $err = join(q{ },
+      "Well this is awkward. I already know someone named $username.",
+      "If that's you, maybe you want <register identity> instead."
```
mmcclimon: Oh yeah, I haven't actually implemented this yet.

lib/Synergy/UserDirectory.pm
```diff
@@ -67,6 +68,47 @@ sub user_by_channel_and_address ($self, $channel_name, $address) {
...
+
+  my $identity_sth = $dbh->prepare('SELECT * FROM user_identities');
+  $identity_sth->execute;
+
+  while (my $row = $identity_sth->fetchrow_hashref) {
+    my $username = $row->{username};
+    my $user = $users{$username};
+
+    unless ($user) {
+      $Logger->log(["Found identity for %s, but no matching user!", $username]);
```
rjbs: …which should be impossible.

lib/Synergy/Channel/Pushover.pm
```diff
@@ -29,15 +29,15 @@ sub http_post {
 }
 
 sub send_message_to_user ($self, $user, $text, $alts = {}) {
-  unless ($user->identities->{ $self->name }) {
+  unless ($user->has_identity_for($self->name)) {
```
rjbs: probably get `$where` first and then check whether it's defined, since identity_for is maybe not fatal? maybe it is, I didn't look…

rjbs:
I have merged this.

PR #98, opened 2019-03-19
UserDirectory: allow "set my nicknames to + add" and "to - remove"

rjbs:
This is janky, but is it better than having to reenter everything?

### DIFF
--- a/lib/Synergy/UserDirectory.pm
+++ b/lib/Synergy/UserDirectory.pm
@@ -16,7 +16,7 @@ use Path::Tiny;
 use Synergy::User;
 use Synergy::Util qw(known_alphabets read_config_file);
 use Synergy::Logger '$Logger';
-use List::Util qw(first shuffle all);
+use List::Util qw(first shuffle all uniq);
 use DateTime;
 use Defined::KV;
 use Try::Tiny;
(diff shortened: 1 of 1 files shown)


lib/Synergy/UserDirectory.pm
```diff
@@ -261,6 +261,23 @@ __PACKAGE__->add_preference(
...
+      return (undef, "no nicknames provided");
+    }
+
+    if ($names[0] eq 'add') {
+      @names = uniq(@names, $event->from_user->nicknames);
+    } elsif ($names[0] eq 'remove') {
+      my %remove = map {; $_ => 1 } @names;
+      @names = grep {; ! $remove{$_} } $event->from_user->nickname;
+    }
+
```
wolfsage: `if ($names[0] eq 'add')` and `elsif ($names[0] eq 'remove')` - I think both of these are supposed to be `$mode`

lib/Synergy/UserDirectory.pm
```diff
@@ -261,6 +261,23 @@ __PACKAGE__->add_preference(
...
+    unless (@names) {
+      return (undef, "no nicknames provided");
+    }
+
+    if ($names[0] eq 'add') {
+      @names = uniq(@names, $event->from_user->nicknames);
+    } elsif ($names[0] eq 'remove') {
+      my %remove = map {; $_ => 1 } @names;
+      @names = grep {; ! $remove{$_} } $event->from_user->nickname;
```
wolfsage: This should probably be `->nicknames`; not `->nickname`

lib/Synergy/UserDirectory.pm
```diff
@@ -261,6 +261,23 @@ __PACKAGE__->add_preference(
                 grep { length $_ }
                 split /\s*,\s*/, $value;
 
+    my $mode = $names[0] eq '+' ? 'add'
+             : $names[0] eq '-' ? 'remove'
+             :                    'set';
+
+    shift @names if $mode ne 'set';
```
wolfsage: So I guess the usage is `set my user.nicknames to + foo,bar` but if I say `set my user.nicknames to +foo,bar` I'll end up with `+foo` and bar` as nicknames? I think the +/- matching should be smarter

PR #99, opened 2019-03-26
WIP: LPC with Futures

rjbs:


lib/Synergy/LPC.pm
```diff
@@ -99,45 +81,52 @@ has http_post_callback => (
...
+  my $res_f = $self->http_get_raw(
     $uri,
     @arg,
+    async => 1,
     Authorization => $self->auth_token,
   );
 
-  return $self->_http_failure($http_res) unless $http_res->is_success;
+  $res_f->then(sub ($res) {
+    $self->log([
```
wolfsage: You're always logging an "error" here even if there isn't one

lib/Synergy/LPC.pm
```diff
@@ -99,45 +81,52 @@ has http_post_callback => (
...
+  my $res_f = $self->http_post_raw(
     $uri,
     @arg,
+    async => 1,
     Authorization => $self->auth_token,
   );
 
-  return $self->_http_failure($http_res) unless $http_res->is_success;
+  $res_f->then(sub ($res) {
+    $self->log([
```
wolfsage: Same here

lib/Synergy/LPC.pm
```diff
@@ -243,34 +231,32 @@ sub track_time ($self, $arg) {
...
+  $res_f->then_with_f(sub ($f, $task) {
+    $self->log([ "response from track_time: %s", $task ]);
 
-  $self->log([ "response from track_time: %s", $res->payload ]);
+    my ($assignment) = grep {; $_->{person_id} == $arg->{member_id} }
+                       $task->{assignments}->@*;
 
-  my ($assignment) = grep {; $_->{person_id} == $arg->{member_id} }
-                     $task->{assignments}->@*;
+    die "WHERE IS MY ASSIGNMENT" unless $assignment;
```
wolfsage: NO HOMEWORK THIS WEEKEND. DON'T RUIN IT FOR THE REST OF THE CLASS RIK

lib/Synergy/LPC.pm
```diff
@@ -243,34 +231,32 @@ sub track_time ($self, $arg) {
...
 
-  die "WHERE IS MY ASSIGNMENT" unless $assignment;
+    if ($arg->{done} xor $assignment->{is_done}) {
+      $f = $f->then(
```
wolfsage: You're just passing arguments to $f->then here. How is this making an API call to update_assignment?

lib/Synergy/LPC.pm
```diff
@@ -299,42 +285,42 @@ sub create_todo_item ($self, $todo) {
...
 sub iteration_relative_to_current ($self, $delta_n) {
   my $helper = LPC::IterationHelper->new({ lpc => $self });
 
-  my $iter    = $helper->iteration_relative_to_current($delta_n);
-  my $pkg_res = $helper->package_for_iteration_number($iter->{number});
+  my $iter  = $helper->iteration_relative_to_current($delta_n);
+  my $pkg_f = $helper->package_for_iteration_number($iter->{number});
 
-  return unless $pkg_res->is_success;
+  return unless $pkg_f->is_success;
```
wolfsage: I think this needs to be `$pkg_f->await->is_done` like above, Future's don't have an `is_success`

lib/Synergy/Reactor/LiquidPlanner.pm
```diff
@@ -1521,16 +1521,14 @@ sub _execute_task_plan ($self, $event, $plan, $error) {
...
-  return unless $res->is_success;
-
-  # What does this mean?  Copied and pasted. -- rjbs, 2018-06-16
-  return unless $res->payload->{start};
-
-  $self->set_last_lp_timer_task_id_for_user($user, $task->{id});
-  return 1;
+  $f->then(sub ($data) {
+    return Future->fail("new timer was not started") unless $data->{start};
+    $self->set_last_lp_timer_task_id_for_user($user, $task->{id});
```
wolfsage: Closing over `$task` feels a little weird but I couldn't say why. I don't think there's any other way to get it here without extreme contortion of futures...

lib/Synergy/Reactor/LiquidPlanner.pm
```diff
@@ -2056,6 +2054,7 @@ sub _do_search ($self, $event, $search, $orig_error = {}) {
     );
   }
 
+  # XXX LPC
```
wolfsage: ?

lib/Synergy/LPC.pm
rjbs: Fixed, thanks.

lib/Synergy/LPC.pm
rjbs: I LIKE HOMEWORK OKAY??

lib/Synergy/LPC.pm
rjbs: This is a big dumb bug.

rjbs: Fixed.

lib/Synergy/LPC.pm
rjbs: Yup, thanks.

lib/Synergy/Reactor/LiquidPlanner.pm
rjbs: I'm not bothered yet.  Maybe later we'll use closure generators.

lib/Synergy/Reactor/LiquidPlanner.pm
rjbs: These comments are notes to me to find and update this code, which doesn't set an $lpc variable, to use futures.  They'll go away when I'm done.

PR #101, opened 2019-06-01
Hub: add http_delete method

robn:
And adjusting http_request to use do_request directly, since
Net::Async::HTTP doesn't provide a DELETE shortcut method.


robn:
Pushed to master 192b3c7.

PR #102, opened 2019-06-01
Slack file upload

robn:
Method to send a file to a user in a private message. It's .. what it is. I could have made api_call a bit more generic but I dunno, this is small and isolated and makes sure nothing else break.

Also a line to ignore Slack events that couldn't be converted to Synergy events, such as the "file uploaded" response event, which looks like a message but is missing a ton fields, including those needed to figure out who the message came from. So `synergy_event_from_slack_event` returns undef, and this makes sure we handle it.

### DIFF
--- a/lib/Synergy/Channel/Slack.pm
+++ b/lib/Synergy/Channel/Slack.pm
@@ -371,6 +371,11 @@ sub maybe_delete_reply ($self, $slack_event) {
   $self->delete_reply($orig_ts);
 }
 
+sub send_file_to_user ($self, $user, $filename, $content) {
+  my $where = $self->slack->dm_channel_for_user($user, $self);
+  return $self->slack->send_file($where, $filename, $content);
+}
+
 sub _uri_from_event ($self, $event) {
   my $channel = $event->transport_data->{channel};
 
--- a/lib/Synergy/Channel/Slack.pm
+++ b/lib/Synergy/Channel/Slack.pm
@@ -143,6 +143,8 @@ sub start ($self) {
     return if $self->slack->username($slack_event->{user}) eq 'synergy';
 
     my $event = $self->synergy_event_from_slack_event($slack_event);
+    return unless $event;
+
     $self->hub->handle_event($event);
   };
 }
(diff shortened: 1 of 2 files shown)


lib/Synergy/Channel/Slack.pm
```diff
@@ -143,6 +143,8 @@ sub start ($self) {
     return if $self->slack->username($slack_event->{user}) eq 'synergy';
 
     my $event = $self->synergy_event_from_slack_event($slack_event);
+    return unless $event;
```
mmcclimon: I'd probably log something here (at least the event type), so we can maybe figure out what we don't handle and deal with them later if we want!

lib/Synergy/External/Slack.pm
```diff
@@ -263,6 +263,34 @@ sub _send_rich_text ($self, $channel, $rich) {
...
 }
 
+sub send_file ($self, $channel, $filename, $content) {
+  my $post_args = [ # arrayref for form data
+    channels => [$channel],
+    filename => $filename,
+    content  => $content,
+  ];
+
+  my $http_future = Future->wrap($self->hub->http->POST(
```
mmcclimon: You don't really need the Future->wrap here, because `$http->POST` returns one anyway. (No harm in having it, though!)

lib/Synergy/External/Slack.pm
```diff
@@ -263,6 +263,34 @@ sub _send_rich_text ($self, $channel, $rich) {
...
+    URI->new($self->_api_url('files.upload')),
+    $post_args,
+    content_type => 'application/x-www-form-urlencoded',
```
mmcclimon: Sigh, Slack.

lib/Synergy/External/Slack.pm
robn: I just copied `api_call`. It seemed weird to me too but I didn't want to spend the time seeing if there was some deeper reason.

lib/Synergy/Channel/Slack.pm
robn: Version I pushed to master just now does this. Dunno how much help it'll be but we'll see.

robn:
Pushed to master a4d7908 and a4d7908.

PR #106, opened 2019-07-26
WIP: initial Github reactor

mmcclimon:
This is just like the Gitlab reactor for now, and needs a bunch of
rework and probably factoring out common stuff. Also there is no
authentication yet.


wolfsage:
Seems good (authentication)?

mmcclimon:
I'm gonna be bold and say this is no good as is, and if a GitHub reactor appears in the future it won't be based on this code.

PR #107, opened 2019-08-13
Upgrade: Only check reactors/channels specified by config

wolfsage:
That way if new modules are added (but not used on the running
instance of synergy) an upgrade will still (rightly) succeed.


rjbs:
5 yr later, I am ready to review, if rebased.

rjbs (changes requested):
needs rebase

PR #111, opened 2019-08-22
yelling: don't complain about URLs

robn:
Sometimes people on the internet say very foolish things and we want to
yell about them. Synergy should not discourage us from our violent mockery.

It was until I was done that I realised that allowing literally any URI is probably too permissive, and a simple `https?:` probably would have been enough. But I still like the word splitting business, easier to do start/end of word that way, and well, here it is.


mmcclimon:
Hey thanks! Merged and then converted to a `for` loop. 😉

PR #112, opened 2019-08-23
VO: manage a Slack usergroup

mmcclimon:
This copies Bort's code for doing the same. This isn't ideal because the
slack method to update user groups requires a privileged API token; this
commit also includes the plumbing for that.

Unlike Bort, this checks oncall on a timer, every 2.5 minutes. That's
because the current VO API for checking oncall no longer includes the
next scheduled shift change, so figuring out when it changes
is...non-trivial, and involves munging a bunch of schedules and stuff.


mmcclimon:
I cherry-picked this, fixed the bugs, and deployed it today.

PR #114, opened 2019-09-06
LP: only nag for overlong timers three times

mmcclimon:
Matthew reports that he was awoken in the middle of the night to 15
texts from Synergy about a timer he left running. That's not great...so
only do so three times, then give up.

### DIFF
--- a/lib/Synergy/Reactor/LiquidPlanner.pm
+++ b/lib/Synergy/Reactor/LiquidPlanner.pm
@@ -1071,6 +1071,11 @@ sub nag ($self, $timer, @) {
           next USER;
         }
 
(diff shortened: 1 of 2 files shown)


lib/Synergy/Reactor/LiquidPlanner.pm
```diff
@@ -1113,7 +1119,6 @@ sub nag ($self, $timer, @) {
           next USER;
         }
         $level = $last_nag->{level} + 1;
-        $Logger->log("$username: setting nag level to $level");
```
wolfsage: Why are we removing this?

rjbs:
LiquidPlanner is dead.

PR #116, opened 2019-10-23
update timer commit TIME to support single digits

tjwds:
Though the comment reads that timer commit TIME 1h should be supported, and it seems to me as though the timer override optionally will grab a group for the floating point portion of a command, the regex did not allow for that portion to be optional.  Because of this, timer commit TIME 1h would commit the timer and add the comment "TIME 1h"

This should _hopefully_ fix this, but I haven't tested it!


mmcclimon:
Applied, thanks!

PR #117, opened 2019-10-23
Have 'spent <...> on' and 'commit TIME <...>' use same parser

wolfsage:
That way we get consistent behaviour, ish, for durations.

This also fixes 'commit TIME 1m' from being 60 hours instead of
60 seconds.


rjbs:
LiquidPlanner is dead.

Issue #123, opened 2020-03-31
InABox reactor crashes if it can't find a snapshot


Davis-A:
This is running master with a small diff to the regex to force i failure to find a snapshot.

````
-        grep { $_->{name} =~ m/^fminabox-/ }
+        grep { $_->{name} =~ m/^zzfminabox-/ }                                                                                                                                                                                                                                          
`````

User requests 'box create' and gets back
>Creating box, this will take a minute or two.
>My inabox reactor crashed while handling your message.  Sorry!

Console output: 
````
adavis@adavis-lenovo:~/src/box-version-Synergy$ carton exec perl bin/synergy -c /home/adavis/src/box-version-Synergy/config.json
LOG | listening on port 8118
LOG | registered /metrics
LOG | Connected to Slack!
LOG | Slack group conversations loaded
LOG | Slack dm channels loaded
LOG | Slack channels loaded
LOG | Users loaded
LOG | message event from slack/u:adavis: box create
LOG | Found SSH key: 24704070 (fminabox)
LOG | fminabox snapshot not found?!
LOG | error with box listener on inabox: Can't use an undefined value as a HASH reference at lib/Synergy/Reactor/InABox.pm line 131.
````

Code path enters the else branch then returns undef.
````
 if ($snapshot) {
        $Logger->log([ "Found snapshot: %s (%s)", $snapshot->{id}, $snapshot->{name} ]);
      }
      else {
        $Logger->log([ "fminabox snapshot not found?!" ]);
      }
      Future->done($snapshot);
    }
````

So then we try to get the id of undef.  And never reach the check.  
````
  my ($snapshot_id, $ssh_key_id) = Future->wait_all(
    $self->_get_snapshot,
    $self->_get_ssh_key,
  )->then;
  (
    sub (@futures) {
      Future->done(map { $_->get->{id} } @futures)
    }
  )->get;

  unless ($snapshot_id && $ssh_key_id) {
    $event->error_reply("Couldn't find snapshot or SSH key, can't create box. Try again.");
    return;
  }
````
My thought is if we get a value back, it will have an id.  We should put the check in the `then` that what we got back was defined and then get the id.  I'm not sure what that looks like though.

mmcclimon:
Oh right, this is actually a problem with `->_get_snapshot`, which can return undef. (It should probably fail the future instead.)

In the meantime, you can work around this (at least the crash) with this diff:

```diff
-      Future->done(map { $_->get->{id} } @futures)
[...]

PR #124, opened 2020-04-01
Inabox new prefs datacentre and version

Davis-A:
This adds new preferences to the inabox reactor.  It also adds the workaround for the reactor crashing when a snapshot isn't found as detailed in #123 

A user can only set a preference based on jessie and buster, but any string can be put in for `/version <foo>` so that if something experimental is being worked on it can be namespaced without needing any configuration change.  Eg `box create /version my-experiment`

The data centre preference will default to the existing behaviour of picking a DC based on timezone, unless a preference has been set.  

At the moment there is no checking of the data centre.  Michael suggested config for this, but i don't think that's the right fit.  For example I recently added Singapore to DC's we upload the nightly snapshot to, but I have no idea what DC's Calvin uploads his WIP buster builds to.  

Ideally i'd like to have access to the snapshot information which lists its available datacentres for that snapshot.  So if their region doesn't match one of the available they get an appropriate message like 
><snapshot-name> is not available in <region> but is available in  <sfo2, sgp1, ... etc>

That will require deeper thinking on how to fix the no snapshot crash bug so the snapshot object isn't thrown away after the id has been extracted.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -145,7 +145,8 @@ sub _handle_create ($self, $event, @args) {
     $self->_get_ssh_key,
   )->then(
     sub (@futures) {
-      Future->done(map { $_->get->{id} } @futures)
+      my @got = grep {; defined } map {; $_->get } @futures;
+      return Future->done(map { $_->{id} } @got);
     }
   )->get;
 
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -628,8 +628,7 @@ sub _update_dns_for_user ($self, $user, $ip) {
 __PACKAGE__->add_preference(
   name      => 'version',
   validator => sub ($self, $value, @) {
-    return (undef, 'version must be jessie or buster') if $value ne 'buster' && $value ne 'jessie';
-
+    return (undef, 'version must be jessie or buster') unless grep { lc $value eq $_ } qw(jessie buster);
     return lc $value;
   },
   default   => 'jessie',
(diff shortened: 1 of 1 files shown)


mmcclimon:
Merged, thanks! I will probably tweak a few minor things here and there, but nothing worth bugging you about.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -493,7 +511,7 @@ sub _get_snapshot ($self) {
       my $data = decode_json($res->content);
       my ($snapshot) =
         sort { $b->{name} cmp $a->{name} }
-        grep { $_->{name} =~ m/^fminabox-/ }
+        grep { $_->{name} =~ m/^fminabox-$version/ }
```
mmcclimon, line 514: The one meaningful change I made  was here

```diff
-         grep { $_->{name} =~ m/^fminabox-$version/ }
+         grep { $_->{name} =~ m/^fminabox-\Q$version\E/ }
```

\Q and \E "quote" the stuff in between, which is safer in case $version contains regex metacharacters (like `.` or `*`). We happen to know that it won't here (it will always be "buster" or "jessie"), but good practice in general!

Davis-A: >We happen to know that it won't here (it will always be "buster" or "jessie"), but good practice in general!

We actually don't know it will always be buster or jessie.  A preference is limited but any string can be used with the `/version` flag

From the original post:
>but any string can be put in for /version <foo> so that if something experimental is being worked on it can be namespaced without needing any configuration change. Eg box create /version my-experiment

So that's a meaningful change to fix.  Thank you for enlightening me on that, i'll keep that in mind in the future.

PR #125, opened 2020-04-02
Inabox reactor refactor

mmcclimon:
Reduce copy/paste, convert to fully async, add a test.


robn:
I spent five minutes skimming this, and it looks really great.

PR #126, opened 2020-04-06
Victorops: ack-all, resolve-all, resolve-mine

Davis-A:
`ack-all` was taken from Michael's fork.
This allows a user to ack all current incidents.

`resolve all` resolves all current acked incidents.
`resolve mine` resolves all current acked incidents that the user acked.

My gripes:  
I don't like how i've handled the difference between mine vs all in the subroutine signature.  It's rather opaque for someone looking at a handler and seeing a 0 or 1 passed in to `_resolve_acked()`

### DIFF
--- a/lib/Synergy/Reactor/VictorOps.pm
+++ b/lib/Synergy/Reactor/VictorOps.pm
@@ -134,6 +134,20 @@ EOH
       method    => 'handle_ack_all',
       predicate => sub ($self, $e) { $e->was_targeted && $e->text =~ /^ack all\s*$/i },
     },
+    {
+      name      => 'resolve-all',
+      method    => 'handle_resolve_all',
+      predicate => sub ($self, $e) {
+        return unless $e->was_targeted;
+        return $e->text =~ m{^resolve\s+all\s*$}i },
+    },
+    {
+      name      => 'resolve-mine',
+      method    => 'handle_resolve_mine',
+      predicate => sub ($self, $e) {
+        return unless $e->was_targeted;
+        return $e->text =~ m{^resolve\s+mine\s*$}i },
+    },
   );
 }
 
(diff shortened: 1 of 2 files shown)


Davis-A:
Possible changes/thoughts i'm having.  rename 'resolve all' to 'resolve acked'

Davis-A:
I also believe i need to move the code to get all the acked incidents into its own method.  

I want to add outstanding unresolved tasks to a users evening report.  Which requires a lot of the same code from `_resolve_acked`

What i'm after at this stage is feedback on how i've written it, any gripes etc

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -222,6 +241,85 @@ sub handle_maint_query ($self, $event) {
...
+      my $noun = $n_acked == 1 ? 'incident' : 'incidents';
+      $event->reply("Successfully resolved $n_acked $noun");
+    })
+    ->else(sub {
+      $event->reply("Something went wrong resolving incidents. Sorry!");
+    })
+    ->retain;
+}
+
+sub _resolve_acked($self, $username, $mine) {
```
mmcclimon, line 270: I agree that `1, 0` here for $mine is weird. I'd probably make it a string, so that you'd say `$self->_resolve_acked($event->from_user->username, 'all')` (or `'mine'`, though I might like `'own'` better, but you can paint this shed however you like).

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -222,6 +241,85 @@ sub handle_maint_query ($self, $event) {
...
+    my $data = decode_json($res->content);
+    my @acked =
+      map { $_->{currentPhase} eq 'ACKED' ? $_ : () }
+      $data->{incidents}->@*;
+
+    return Future->done(0) unless @acked;
+
+    if ($mine) {
+      @acked =
+        map {$_->{transitions}[-1]->{by} eq $self->vo_from_username($username) ? $_ : ()}
```
mmcclimon, line 290: It is more idiomatic to use `grep` rather than `map` here, which would be 

```perl
@acked = grep { $_->{transitions}[-1]->{by} eq $self->vo_from_username($username) } @acked
```

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -222,6 +241,85 @@ sub handle_maint_query ($self, $event) {
...
+    $self->_vo_api_headers,
+    async => 1,
+  )->then (sub ($res) {
+    unless ($res->is_success) {
+      $Logger->log("VO: get incidents failed: ".$res->as_string);
+      return Future->fail('get incidents');
+    }
+
+    my $data = decode_json($res->content);
+    my @acked =
```
mmcclimon, line 282: Same here: `grep` is for filtering out a list, `map` for transforming them. If you ever do `map { $some_condition ? $_ : () } @list`, that's better written as `grep { $some_condition } @list`.  (Mapping the empty list is also a little strange, but I know that code was existing; I'd probably write that as a map followed by a grep, or vice versa).

So here: 
```perl
my @acked = grep { $_->{currentPhase} eq 'ACKED' } $data->{incidents}->@*;
```

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -222,6 +241,85 @@ sub handle_maint_query ($self, $event) {
...
+
+    return Future->done(0) unless @acked;
+
+    if ($mine) {
+      @acked =
+        map {$_->{transitions}[-1]->{by} eq $self->vo_from_username($username) ? $_ : ()}
+        @acked
+    }
+
+    my @unresolved =
```
mmcclimon, line 294: This is weird @acked at this point should not have any undefs in it! (If it does, that probably indicates something weird a little earlier. That means this could be 

```perl
my @unresolved = map { $_->{incidentNumber} } @acked
```

Davis-A: This was to pull out the incident numbers.  As up till this point i have the whole incident object.  

````
    $self->hub->http_patch(
      $self->_vo_api_endpoint('/incidents/resolve'),
      $self->_vo_api_headers,
      async => 1,
      Content_Type => 'application/json',
      Content => encode_json({
        userName => $self->vo_from_username($username),
        incidentNames => \@unresolved,
      }),
    );
````

Is there a way i can do something like `incidentNames =>\@unresolved->{incidentNumber}` to avoid pulling the incidentNumbers out earlier?

Comment (mmcclimon) [reply lib/Synergy/Reactor/VictorOps.pm:278]: I was unclear, sorry. I was objecting to the `map { $_ ? ... : () }` construction. By this point, `@acked` shouldn't have anything undef in it, so you don't need the ternary, and including it is confusing. If you just get rid of that, the code is otherwise fine:

```perl
my @unresolved = map { $_->{incidentNumber} } @acked
```

PR #127, opened 2020-04-09
WIP: VO: Refactor resolving incidents

Davis-A:
Still a work in progress.

Could i get some feedback on the `sub _resolve_incidents()` which could be used to unify resolving all incidents ie replacing `_resolve_acked()` which does the same thing, except it only filters for acked and who acked.  Whereas resolve incidents can filter on timestamp, acked vs unackked as well.  At this stage i haven't ripped it out yet, no point till the final look or `_resolve_incidents()` is up to snuff.

The other part is the `/resolve` to demaint.  It's awkward because VO doesn't return the timestamp when you end maint, just an empty transformations [ ].  

Tests also need to be updated etc.  

Possible ideas of altering `_resolve_acked()`, given there are now quite a few arguments is replace them with a hash.

### DIFF
--- a/lib/Synergy/Reactor/VictorOps.pm
+++ b/lib/Synergy/Reactor/VictorOps.pm
@@ -14,6 +14,7 @@ use JSON::MaybeXS;
 use List::Util qw(first);
 use DateTimeX::Format::Ago;
 use Synergy::Logger '$Logger';
+use Date::Parse;
 
 has alert_endpoint_uri => (
   is => 'ro',
@@ -119,9 +120,9 @@ EOH
       method    => 'handle_maint_end',
       predicate => sub ($self, $e) {
         return unless $e->was_targeted;
-        return 1 if $e->text =~ /^maint\s+(end|stop)\s*$/i;
-        return 1 if $e->text =~ /^unmaint\s*$/i;
-        return 1 if $e->text =~ /^demaint\s*$/i;
+        return 1 if $e->text =~ /^maint\s+(end|stop)\b/i;
+        return 1 if $e->text =~ /^unmaint\b/i;
+        return 1 if $e->text =~ /^demaint\b/i;
       },
     },
     {
@@ -140,6 +141,13 @@ EOH
       predicate => sub ($self, $e) {
         return unless $e->was_targeted;
         return $e->text =~ m{^resolve\s+all\s*$}i },
+    },
+    {
+      name      => 'resolve-acked',
+      method    => 'handle_resolve_acked',
+      predicate => sub ($self, $e) {
+        return unless $e->was_targeted;
+        return $e->text =~ m{^resolve\s+acked\s*$}i },
     },
     {
       name      => 'resolve-mine',
@@ -256,43 +264,27 @@ sub handle_maint_query ($self, $event) {
 }
 
 sub handle_resolve_mine ($self, $event) {
-  return $self->_resolve_acked($event, 'mine');
+  $event->mark_handled;
+  $self->_resolve_incidents($event, {
+    type => 'acked',
+    whose => 'own'
+  })->retain;
 }
 
 sub handle_resolve_all ($self, $event) {
-  return $self->_resolve_acked($event, 'all');
+  $event->mark_handled;
+  $self->_resolve_incidents($event, {
+    type => 'all',
+    whose => 'all'
+  })->retain;
 }
 
-sub _resolve_acked($self, $event, $whose) {
+sub handle_resolve_acked ($self, $event) {
   $event->mark_handled;
-
-  $self->_vo_request(GET => '/incidents')
-    ->then (sub ($data) {
-      my @acked = grep {; $_->{currentPhase} eq 'ACKED' } $data->{incidents}->@*;
-      return Future->done({ results => []}) unless @acked;
-
-      my $vo_username = $self->vo_from_username($event->from_user->username);
-
-      if ($whose eq 'mine') {
-        @acked = grep {; $_->{transitions}[-1]->{by} eq $vo_username } @acked;
-      }
-
-      my @unresolved = map {; $_->{incidentNumber} } @acked;
-
-      return $self->_vo_request(PATCH => "/incidents/resolve", {
-        userName => $vo_username,
-        incidentNames => \@unresolved,
-      });
-    })->then(sub ($data) {
-      my $n = $data->{results}->@*;
-      my $noun = $n == 1 ? 'incident' : 'incidents';
-
-      my $exclamation = $whose eq 'all' ? "The board is clear!" : "Phew!";
-
-      $event->reply("Successfully resolved $n $noun. $exclamation");
-    })->else(sub {
-      $event->reply("Something went wrong resolving incidents. Sorry!");
-    })->retain;
+  $self->_resolve_incidents($event, {
+    type => 'acked',
+    whose => 'all'
+  })->retain;
 }
 
 sub _current_oncall_names ($self) {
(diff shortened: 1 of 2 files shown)


Davis-A:
@mmcclimon

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -390,9 +391,44 @@ sub handle_maint_start ($self, $event) {
...
 
+sub _resolve_incidents($self, $event, $type, $since, $whose) {
+  return $self->_vo_request(GET => '/incidents')
+    ->then(sub($data){
+      my $vo_username = $self->vo_from_username($event->from_user->username);
+
+      my @unresolved = grep {; $_->{currentPhase} ne 'RESOLVED' } $data->{incidents}->@*;
+      # filter by time, maint window time unix epoch with milliscond resolution
+      # convert incident time to unix time, maint time is unix epoch with millisecond resolution
+      @unresolved = grep {; (str2time($_->{startTime}) * 1000) > $since } @unresolved unless $since == 0;
```
mmcclimon, line 402: I find this block a bit hard to read, partly because we're reusing the same variable throughout and partly because the conditions are hung way off to the right. I might lean toward writing this as a loop instead; something like

```perl
my @unresolved;

for my $incident ($data->{incidents}->@*) {
  next if $incident->{currentPhase} eq 'RESOLVED';

  if (my $since = $arg->{since}) {
    next unless ($incident->{startTime} / 1000) > $since;
  }

  unless ($type eq 'all') {
    # skip unacked incidents unless we've asked for all
    next if $incident->{currentPhase} eq 'ACKED';
  }

  if ($arg->{whose} eq 'own' && $type ne 'all') {
    next unless $incident->{transitions}[-1]->{by} eq $vo_username;
  }

  push $unresolved, $incident->{incidentNumber};
}
```

(And then at the end, `@unresolved` only contains the incident numbers you're interested in.)

mmcclimon: Also, I think using `str2time` is overkill here: if those are just millisecond epoch stamps, you can get the regular epoch with `$incident->{startTime} / 1000` (or `int($incident->{startTime} / 1000)` if you don't want the fractional part.

...but if the maint windows are _also_ millisecond epoch stamps, there's no need to convert at all and you can just compare them!

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -390,9 +391,44 @@ sub handle_maint_start ($self, $event) {
...
+      $event->reply("Something went wrong resolving incidents. Sorry!");
+    });
+}
+
 sub handle_maint_end ($self, $event) {
   $event->mark_handled;
 
+  my (@args) = split /\s+/, $event->text;
+  my $resolve = grep {; $_ eq '/resolve' } @args;
+  return;
```
mmcclimon, line 430: (errant return, I assume because WIP.)

Davis-A: Yeah, sorry.  I didn't meant to leave that there.  

I was hitting victorops for actual data and returning there to not actually act on anything real

lib/Synergy/Reactor/VictorOps.pm
Davis-A: Re str2time:
Victorops API is stupid.  startTime on on maintenance are millisecond unix epoch.
Incident startTime are this format '2020-04-08T23:20:50Z'

Davis-A:
Ok, newly updated.

I also made a behaviour change i'd need to let everyone know about.

I somewhat disliked "resolve all" only resolved currently acked items.  
So now "resolve all" will resolve all items, and a new listener "resolve acked" will resolve all currently acked incidents.  "resolve mine" will continue to behave the same, only resolving my own acked incidents.  

I still need to jump onto the tests.  But other than that, how does this look?

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -119,9 +120,9 @@ EOH
       method    => 'handle_maint_end',
       predicate => sub ($self, $e) {
         return unless $e->was_targeted;
-        return 1 if $e->text =~ /^maint\s+(end|stop)\s*$/i;
-        return 1 if $e->text =~ /^unmaint\s*$/i;
-        return 1 if $e->text =~ /^demaint\s*$/i;
+        return 1 if $e->text =~ /^maint\s+(end|stop)\b/i;
+        return 1 if $e->text =~ /^unmaint\b/i;
```
mmcclimon, line 124: This is a behavior change (it's probably fine, but): this used to only match the word "unmaint" and now  it will match "unmaint all of the cyrus slots," which is a little weird.

mmcclimon: (Ditto with the rest of these `\b` changes.)

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -409,9 +448,23 @@ sub handle_maint_end ($self, $event) {
...
+      my $timestamp = $maint->[0]->{startedAt};
       my $instance_id = $global_maint->{instanceId};
 
-      return $self->_vo_request(PUT => "/maintenancemode/$instance_id/end");
+      return $self->_vo_request(PUT => "/maintenancemode/$instance_id/end")
+        ->transform(sub ($data) { return ($data, $timestamp) })
+    })
+    ->then(sub ($data, $timestamp) {
+      if ($resolve) {
+        return $self->_resolve_incidents($event, {
```
mmcclimon, line 460: So, I think this won't do exactly what you want (it might, but only by accident). `_resolve_incidents` returns the result of Future->retain. (Which is to say, it's a "final" handler, in that it will always respond, even in the case of failure, and its return value is mostly incidental.) 

Here, you're using the return value as a proper Future, that something else depends on, which may or may not do what you want. Like, if `_resolve_incidents` returns a nullary Future, you'll get a "wrong number of arguments for subroutine" crash when we get to the `then` below.

This is easy to fix, though. You can just change this `then` to:

```perl
    ->then(sub ($data, $timestamp) {
      if ($resolve) {
        $self->_resolve_incidents($event, {
          type => 'all',
          since => $timestamp,
          whose => 'all'
        });
      }

      return Future->done(0);
    })
```

so that either way we know the return value is correct.

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -409,9 +448,23 @@ sub handle_maint_end ($self, $event) {
           "You'll need to go and sort it out in the VO web UI.");
       }
 
+
+      my $timestamp = $maint->[0]->{startedAt};
       my $instance_id = $global_maint->{instanceId};
 
-      return $self->_vo_request(PUT => "/maintenancemode/$instance_id/end");
+      return $self->_vo_request(PUT => "/maintenancemode/$instance_id/end")
+        ->transform(sub ($data) { return ($data, $timestamp) })
```
mmcclimon, line 456: You're already transforming this, and you don't use $data below, so you could just change this to `sub { $timestamp }` if you want.

t/victorops.t
```diff
@@ -190,22 +211,56 @@ subtest 'exit maint' => sub {
...
-# We want to test, for our PATCH requests, that we only request the ones we
-# expect. So here, we return as many successful results as there were
-# incidents. Later, if we make the assertions in the VO reactor stronger
-# (rather than just returning a count of the results without checking their
-# content), this will be easier to extend.
-my $patch_responder = sub ($req) {
-  note("got JSON data: " . $req->content);
-  my $patch_data = decode_json($req->content);
+    # exit maint and /resolve
+  # 55, 56 should resolve
```
mmcclimon, line 216: This indentation is wonky.

mmcclimon: Also, the numbers in this comment don't match the incident numbers in the test.

t/victorops.t
```diff
@@ -190,22 +211,56 @@ subtest 'exit maint' => sub {
...
-  return gen_response(200, { results => \@incidents });
+    send_message('synergy: demaint /resolve', 'alice');
+    like(
+      multiple_message_text(2, 0),
+      qr{Successfully resolved 2 incidents. The board is clear!}i,
+      'demaint /resolve'
+    );
 };
 
+
```
mmcclimon, line 263: (unnecessary whitespace.)

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -389,10 +381,54 @@ sub handle_maint_start ($self, $event) {
 
   return;
 }
+sub _resolve_incidents($self, $event, $args) {
```
mmcclimon, line 384: (This new sub should have an empty line before it.)

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -409,9 +445,22 @@ sub handle_maint_end ($self, $event) {
...
+    })
+    ->then(sub ($timestamp) {
+      if ($resolve) {
+        return $self->_resolve_incidents($event, {
+          type => 'all',
+          since => $timestamp,
+          whose => 'all'
+        });
+      }
+        return Future->done(0);
```
mmcclimon, line 463: (Wonky indentation.)

Davis-A:
Ok fixed the whitespace, sorry about those that's definitely my bad.

>also (more importantly), the tests don't pass.

What tests aren't passing?

````
adavis@adavis-lenovo:~/src/maint-end-resolve-synergy$ carton exec prove t/victorops.t
t/victorops.t .. ok
All tests successful.
Files=1, Tests=6,  2 wallclock secs ( 0.00 usr  0.03 sys +  0.67 cusr  0.29 csys =  0.99 CPU)
Result: PASS
adavis@adavis-lenovo:~/src/maint-end-resolve-synergy$
````

mmcclimon:
The test failure was an issue with Date::Parse, fixed recently (I was on an older version, as was Synergy production). I have merged this, and updated the cpanfile to match. Thanks!

PR #128, opened 2020-05-26
Hub: all http requests are now async

rjbs:
or, well, they return futures, anyway


mmcclimon:
I haven't audited the code for everywhere we do an http call, but I think this is fine, and the places that aren't will shake out quickly, I imagine.

PR #131, opened 2020-09-15
VictorOps: "maint end" always resolves, and 10m more of alerts

rjbs:


rjbs:
This is not ready for merge, because the tests fail now.  I will think about them later.

rjbs:
@Davis-A I will send this to you for review when you've accepted invite. :)

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -521,24 +523,22 @@ sub handle_maint_end ($self, $event) {
...
       # We resolve before demainting because there's a race: if you exit
       # maint, VO immediately sends phone alerts for everything that's active,
       # even if they're going to be resolved in a quarter-second. That's
       # really annoying, so we resolve everything, then tell VO to stop maint,
       # at which point there shouldn't be anything left to buzz.
       return $self->_resolve_incidents($event, {
         type => 'all',
-        since => $timestamp,
-        whose => 'all'
+        since => $timestamp - 600,
```
Davis-A, line 539: `startedAt` timestamp is in Unixtime * 1000 for millisecond resolution hence this is only winding our resolution time back by 0.6 seconds instead of 600.

VO API is poorly documented and inconsistent.  As if you look in _resolve_incidents you'll discover that for an incident the `startTime` (why the heck is it name differently too) is Unixtime....

Davis-A:
Other than being bitten by the VO API timestamps being all over the shop it looks good.

Davis-A (changes requested):
Other than being bitten by the VO API timestamps being all over the shop it looks good.

Davis-A (approved):

Davis-A:
So the tests only failed after my commit on top to fix the 10 minute window Rik wants.  
This led me to discover there was a bug in the test.  It was setting the maint `startedAt` field to 86400.  `startedAt` is milliseconds since the unix epoch so that was setting at 86.4 seconds past epoch.

So a fixed that, then reworked the demaint /resolve test to include an additional incident before `startedAt`  but within the 10 minute window and adjusted one of the existing incidents to be closer to `startedAt` but outside of the 10 minute window

PR #132, opened 2020-09-23
vo: maint timer periodically notifies oncall we're still in maint

Davis-A:
Sometimes operators forget to demaint.

Create a timer that fires every 30 minutes to @oncall in plumbing and remind them we're in maint.

Not sure if how i've set up the timer etc is the best way to do things.  Especially removing the timer.

### DIFF
--- a/lib/Synergy/Reactor/VictorOps.pm
+++ b/lib/Synergy/Reactor/VictorOps.pm
@@ -255,6 +255,7 @@ EOH
 sub state ($self) {
   return {
     oncall_list => $self->oncall_list,
+    maint_started_by_user => $self->maint_started_by_user,
   };
 }
 
@@ -501,6 +502,7 @@ sub handle_maint_start ($self, $event) {
   })
   ->then(sub ($data) {
     $self->maint_started_by_user($event->from_user->username);
+    $self->save_state;
     $self->_ack_all($event->from_user->username);
   })
   ->then(sub ($nacked) {
--- a/lib/Synergy/Reactor/VictorOps.pm
+++ b/lib/Synergy/Reactor/VictorOps.pm
@@ -92,6 +92,23 @@ has oncall_group_address => (
   isa => 'Str',
 );
 
+has maint_warning_address => (
+  is  => 'ro',
+  isa => 'Str',
+);
+
+has maint_timer_interval => (
+  is => 'ro',
+  isa => 'Int',
+  default => 600,
+);
+
+has maint_started_by_user => (
+  is => 'rw',
+  isa => 'Str',
+  clearer => '_clear_maint_started_by_user',
+);
+
 has oncall_list => (
   is => 'ro',
   isa => 'ArrayRef',
@@ -108,14 +125,21 @@ around '_set_oncall_list' => sub ($orig, $self, @rest) {
 sub start ($self) {
   return unless $self->oncall_channel_name;
 
-  my $timer = IO::Async::Timer::Periodic->new(
+  my $check_oncall_timer = IO::Async::Timer::Periodic->new(
     first_interval => 30,   # don't start immediately
     interval       => 150,
     on_tick        => sub { $self->_check_at_oncall },
   );
 
-  $timer->start;
-  $self->hub->loop->add($timer);
+  $check_oncall_timer->start;
+  $self->hub->loop->add($check_oncall_timer);
+
+  my $maint_warning_timer = IO::Async::Timer::Periodic->new(
+    interval => $self->maint_timer_interval,
+    on_tick  => sub {  $self->_check_long_maint },
+  );
+  $maint_warning_timer->start;
+  $self->hub->loop->add($maint_warning_timer);
 }
 
 sub listener_specs {
@@ -354,6 +378,57 @@ sub handle_resolve_acked ($self, $event) {
   })->retain;
 }
 
+sub _check_long_maint ($self) {
+  my $maint;
+  my $current_time = time();
+  $self->_vo_request(GET => '/maintenancemode')
+  ->then(sub ($data) {
+    $maint = $data->{activeInstances} // [];
+    unless (@$maint) {
+      $self->_clear_maint_started_by_user();
+      return Future->fail('not in maint');
+    }
+    return $self->_vo_request(GET => '/incidents');
+  })->then(sub ($data) {
+    my $is_new_incident;
+
+    for my $incident ($data->{incidents}->@*) {
+      next if $incident->{currentPhase} eq 'RESOLVED';
+
+      my $incident_start_time = str2time($incident->{startTime});
+      my $incident_duration = $current_time - $incident_start_time;
+
+      $is_new_incident = 1, last if $incident_duration < (60 * 25);
+    }
+    return Future->fail('an incident newer than 25 minutes') if $is_new_incident;
+
+    # maint startedAt is unix time * 1000
+    my $maint_start_time = int($maint->[0]->{startedAt} / 1000);
+    my $maint_duration_s = $current_time - $maint_start_time;
+    return Future->fail('maint duration less than 30m') unless $maint_duration_s > (60 * 30);
+
+    return Future->fail('oncall_channel_name undefined') unless $self->oncall_channel_name;
+    return Future->fail('oncall_group_address undefined') unless my $group_address = $self->oncall_group_address;
+    return Future->fail('cannot create channel') unless my $channel = $self->hub->channel_named($self->oncall_channel_name);
+
+    if ($channel && $self->maint_warning_address) {
+      my $maint_duration_m = int($maint_duration_s / 60);
+      my $who = $maint->[0]->{startedBy} eq 'PUBLICAPI' ? $self->maint_started_by_user : $maint->[0]->{startedBy};
+      $channel->send_message(
+        $self->maint_warning_address,
+        "\@oncall Hey, by the way, VictorOps is in maintenance mode. (Started $maint_duration_m minutes ago by $who.)",
+        { slack => "<!subteam^$group_address> Hey, by the way, VictorOps is in maintenance mode. (Started $maint_duration_m minutes ago by $who.)" }
+      );
+    }
+  })
+  ->else(sub ($message, $extra = {}) {
+    return if $message eq 'not in maint';
+    return if $message eq 'an incident newer than 25 minutes';
+    return if $message eq 'maint duration less than 30m';
+    $Logger->log("VO: error _check_long_maint():  $message");
+  })->retain;
+}
+
 sub _current_oncall_names ($self) {
   $self->_vo_request(GET => '/oncall/current')
     ->then(sub ($data) {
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -354,6 +365,28 @@ sub handle_resolve_acked ($self, $event) {
   })->retain;
 }
 
+sub _remove_maint_timer ($self) {
+  my $timer = $self->maint_warning_timer;
+  $self->hub->loop->remove($timer);
+  $self->maint_warning_timer(0);
+}
+
+sub _long_maint_warning ($self, $timer) {
```
Davis-A, line 374: I'm not exactly sure how to direct a message to plumbing and target @oncall. 
This is the only piece of the puzzle unimplemented

mmcclimon: So, we do this in a handfull of places. It involves configuring the address (at-oncall, which we already have) and channel (pound-plumbing, which we don't), then using that. The best place to steal from here would be the LP reactor; search for `triage_address` and `_inform_triage`.

You'll eventually want something like 
```perl
  if ($self->oncall_channel_name && $self->maint_warning_address) {
    $channel->send_message($self->maint_warning_address, $text, $alt);
  }
```

Also, this method should check, before it does anything, if we're still in maint (and if we're not, cancel the timer!). 

Also also, I dislike this method name, because it's not a noun and not a verb, and I'd probably call it `send_long_maint_warning` or something.

mmcclimon (changes requested):
Back to you.

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -216,6 +220,13 @@ EOH
         return unless $e->was_targeted;
         return $e->text =~ m{^resolve\s+mine\s*$}i },
     },
+    {
+      name      => 'maint-timer-stop',
+      method    => 'handle_maint_timer_stop',
+      predicate => sub ($self, $e) {
+        return unless $e->was_targeted;
+        return $e->text =~ m{^maint\s+timer\s+stop\s*$}i },
```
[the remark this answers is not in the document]
Davis-A: My thought was if there is ever a time we don't want it we can turn it off and it doesn't hurt to have it.  But then I guess we can activate maint via the VictorOps webui instead.

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -354,6 +365,28 @@ sub handle_resolve_acked ($self, $event) {
   })->retain;
 }
 
+sub _remove_maint_timer ($self) {
+  my $timer = $self->maint_warning_timer;
+  $self->hub->loop->remove($timer);
+  $self->maint_warning_timer(0);
```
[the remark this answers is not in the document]
Davis-A: I knew this would come up in the review, this was a placeholder till someone told me a better way to handle it.  To get this far I just needed something to check an if condition against and that worked.  
Thanks for telling me about `clearer` 😄

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -418,6 +451,21 @@ sub handle_maint_start ($self, $event) {
...
+    # alert oncall every 30 minutes when in maint
+    my $maint_warning_timer = IO::Async::Timer::Periodic->new(
+      interval => 1800,
+      on_tick  => sub ($timer, @arg) {  $self->_long_maint_warning($timer) },
+    );
+
+    $self->hub->loop->add($maint_warning_timer);
+    $maint_warning_timer->start;
+    $self->maint_warning_timer($maint_warning_timer);
+    return Future->done(0);
```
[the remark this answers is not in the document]
Davis-A: Okie dokie.  I've used that before in the ack all and resolve all patches.  I think I copied that from the Liquid Planner reactor, a great case of I copied the 1/20 of them that was wrong and have done it wrong myself ever since.

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -354,6 +370,36 @@ sub handle_resolve_acked ($self, $event) {
...
 }
 
+sub _remove_maint_timer ($self) {
+  my $timer = $self->maint_warning_timer;
+  return unless $timer;
+  $self->hub->loop->remove($timer);
+  $self->_clear_maint_warning_timer;
+}
+
+sub _send_long_maint_warning ($self, $timer) {
```
Davis-A, line 380: Two things in this method:

**One:**
I'm not sending messages to oncall.  I was hoping to find a way to send a single message to #plumbing targeting \@oncall and i couldn't figure it out, at least without more time.  So i'm wondering what you think, whether it should just send to plumbing, or follow the triage model and ping each oncall member individually in addition to sending to #plumbing?.  

**Two:**
I hate how I implemented this.  I could think of 3 bad ways to implement so I picked one.  What i really think would be easier is do the request then the `then( sub ....)` just assign `$data` to a variable declared before the request, but that also felt icky.  So really I could only think of 3 bad ideas so I picked one, which unfortunately puts the burden on you to tell me how to do it right :(

I wanted to follow the pattern in the liquidplanner reactor which is 

````
return unless $self->triage_channel_name;
return unless my $channel = $self->hub->channel_named($self->triage_channel_name);
[...]
	  
if ($self->triage_channel_name && $self->triage_address) {
  $channel->send_message($self->triage_address, $text, $alt);
}
````

But if we're in this method then the timer exists so we need to clear it if we aren't in maint.  Meaning this check has to exist after the check of maint mode.

Comment (mmcclimon) [reply lib/Synergy/Reactor/VictorOps.pm:380]: btw: you can totally target oncall, using the syntax at https://api.slack.com/reference/surfaces/formatting#mentioning-groups. It'll be something like `<!subteam^S12345>`, where `S12345` is the id of the oncall group.

File: lib/Synergy/Reactor/VictorOps.pm [hunk 6]
```diff
@@ -354,6 +370,36 @@ sub handle_resolve_acked ($self, $event) {
...
+    }
+
+    return Future->done unless $self->oncall_channel_name;
+    return Future->done unless my $channel =
+      $self->hub->channel_named($self->oncall_channel_name);
+
+    if ($channel && $self->maint_warning_address) {
+      $channel->send_message(
+        $self->maint_warning_address,
+        'We are still in maint',
```
mmcclimon, line 396: I'd provide more detail here. I'd want to see something like "@oncall Hey, by the way, VictorOps is in maintenance mode. (Started $n minutes ago by $who.)"  I don't know if we can actually tell who started it (probably not), but knowing how long it's been around is definitely useful.

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -419,6 +465,21 @@ sub handle_maint_start ($self, $event) {
     });
   })
   ->then(sub ($data) {
+    # Check for a bogus timer, possible if maint is disabled elsewhere
+    $self->_remove_maint_timer();
```
mmcclimon, line 469: So (annoying comment to follow):

I think it would be better not to stop and start this timer all the time, but rather have a timer that runs _all_ the time, just like the one that checks for changes in oncall. If I were doing it, I'd it to tick every 10 minutes, and only have it alert if it had been more than 30 minutes since maint started (and at least 25 minutes since the last alert).

This way, we still get the warning if maint is started from not-Synergy. I thought of this because yesterday we got a bunch of alerts while Slack was down, and I have occasionally started maint from my phone. If someone does that and forgets to demaint, we _especially_ want to know, because the original maint wasn't publicly visible.

mmcclimon (changes requested):
With apologies I didn't think of it sooner, I have suggested a slightly different tack which I think will be a better experience and also the implementation will be less fiddly.

lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -354,6 +378,57 @@ sub handle_resolve_acked ($self, $event) {
...
+  })->then(sub ($data) {
+    my $is_new_incident;
+
+    for my $incident ($data->{incidents}->@*) {
+      next if $incident->{currentPhase} eq 'RESOLVED';
+
+      my $incident_start_time = str2time($incident->{startTime});
+      my $incident_duration = $current_time - $incident_start_time;
+
+      $is_new_incident = 1, last if $incident_duration < (60 * 25);
```
mmcclimon, line 401: I'm not sure what this `$incident_duration` stuff is even doing. I think an incident is an alert, and the point of this block is that we don't send the warning if there's been an alert in the last 25 minutes?

A few things: 
a) `$is_new_incident = 1, last` reads weirdly to me, and I'd split it into different statements.
b) But that's no so important, because I'm not sure that's a great idea anyway. Say there's some alert fire, and we enter maint. That incident ends in 15 minutes, but maint was not resolved. Some time later, some unrelated thing fires. I think we still want the "hey we're in maint" warning, because that second alert might not have been noticed in the meantime (by virtue of the fact that we're in maint, which we might have forgotten)!

mmcclimon (changes requested):
This looks pretty good. I have made some comments, and will make some more now:

1. This will alert us every five minutes (the timer fires every 600s, and warns us if maint duration is more than 30 minutes). I think that's probably overkill. Having the timer run every 5 minutes is fine, but I would save the last-warning time into state and check that at the top of the timer tick routine.

2. The state (who started maint, and I suggest also the last time we notified about maint) will not survive a restart, and it should. That's as simple as putting that stuff into `sub state`, and making sure that `->save_state` gets called after they're set.

3. I would reconsider the current behavior of suppressing the notification if there's been an alert recently. This might turn out to be a bad idea, but I'd rather synergy be too noisy and tune-downable rather than have us miss the warnings because rblcheck or whatever happened to fire at an inopportune time.

mmcclimon:
Oh also: I would always prefer more commits for substantial changes rather than amending the same one over and over. That gives you an opportunity to write a bit about why the changes are necessary, rather than having them lost to the sands of time in Github PR comments.

lib/Synergy/Reactor/VictorOps.pm
Davis-A: a) i've seen that pattern in a dozen of our cli tools, so i thought it was acceptable.

b) Incident is VO nomenclature for an alert.  
Totally agree it is strange.  But it was based on your feedback.

>If I were doing it, I'd it to tick every 10 minutes, and only have it alert if it had been more than 30 minutes since maint started (and at least 25 minutes since the last alert).

Something lost in translation I guess.  How was I supposed to interpret the latter part in the ( ) ?

mmcclimon: a) I only recall having seen it in `print($usage->text), exit if $opt->help`, which is an idiom by itself and basically the only place I ever use the comma operator.

b) Aha, this was totally clear to me and I did not express it well even a little. I meant: if it had been more than 30 minutes since maint started (and at least 25 minutes since the last time Synergy notified us about ongoing maint). That is: "alert" in parens was "the synergy long-maint alert," but I can see now that that is not at all clear. Sorry!

mmcclimon: Which is to say: I think we should get this long-maint warning every 30m or so while we're in maint, but that involves a bit of housekeeping to make sure you don't send it on every tick.

Davis-A:
I think i've got everything!

mmcclimon:
@Davis-A Thanks! I have added two fiddly commits on top of this and merged.

PR #133, opened 2020-10-09
inabox: add support for multiple boxes

robn:
Extending `/version` and adding `/tag`, to be able to create multiple boxes, each with different names and/or versions.

After deploying, I will rename all existing boxes to be <username>-jessie, which should make the change pretty much invisible to people that only want/need one box.

I think it could be made a lot nicer, particularly when it comes to targetting existing boxes. On the other hand, this should mostly be unuseful except outside of the OS upgrade in progress and certain weird situations, and a little friction maybe isn't bad because we don't actually want people creating multiples most of the time.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -155,34 +155,43 @@ sub _do_request ($self, $method, $endpoint, $data = undef) {
   });
 }
 
+sub _determine_version_and_tag ($self, $event, $switches) {
+  # this convoluted mess is about figuring out:
+  # - the version, by request or from prefs or implied
+  # - the tag, by request or from the version
+  # - if this is the "default" box, which sets the "$username.box" DNS name
+  my $default_version = $self->get_user_preference($event->from_user, 'version');
+  my ($version, $tag) = $switches->@{qw(version tag)};
+  $version //= $default_version;
+  $tag //= $version;
+
+  # XXX check version and tag valid
+
+  return ($version, $tag);
+}
+
 sub handle_status ($self, $event, $switches) {
-  $self->_get_droplet_for($event->from_user)
-    ->then(sub ($droplet = undef) {
-      if ($droplet) {
-        return $event->reply("Your box: " . $self->_format_droplet($droplet));
+  $self->_get_droplets_for($event->from_user)
+    ->then(sub ($droplets = undef) {
+      if (@$droplets) {
+        return $event->reply(join "\n",
+          "Your boxes: ",
+          map { $self->_format_droplet($_) } @$droplets,
+        );
       }
 
-      return $event->reply("You don't seem to have a box.");
+      return $event->reply("You don't seem to have any boxes.");
     });
 }
 
 sub handle_create ($self, $event, $switches) {
-  # do sanity checks before HTTP requests
-  my $version = delete $switches->{version}
-             // $self->get_user_preference($event->from_user, 'version');
+  my ($version, $tag) = $self->_determine_version_and_tag($event, $switches);
 
-  if (keys %$switches) {
-    return Future->fail(
-      'error: syntax is "box create [ /version <version> ]"',
-      'stop-processing',
-    );
-  }
-
-  return $self->_get_droplet_for($event->from_user)
+  return $self->_get_droplet_for($event->from_user, $tag)
     ->then(sub ($maybe_droplet) {
       if ($maybe_droplet) {
         return Future->fail(
-          "You already have a box: " . $self->_format_droplet($maybe_droplet),
+          "This box already exists: " . $self->_format_droplet($maybe_droplet),
           'stop-processing'
         );
       }
@@ -202,11 +211,13 @@ sub handle_create ($self, $event, $switches) {
       });
     })
     ->then(sub ($snapshot, $ssh_key) {
+      my $name = $self->_box_name_for($event->from_user, $tag);
+
       my $region = $self->_region_for_user($event->from_user);
-      $event->reply("Creating $version box in $region, this will take a minute or two.");
+      $event->reply("Creating $name in $region, this will take a minute or two.");
 
       my %droplet_create_args = (
-        name     => $self->_box_name_for($event->from_user),
+        name     => $name,
         region   => $region,
         size     => 's-4vcpu-8gb',
         image    => $snapshot->{id},
@@ -243,7 +254,7 @@ sub handle_create ($self, $event, $switches) {
 
       return Future->done;
     })
-    ->then(sub { $self->_get_droplet_for($event->from_user) })
+    ->then(sub { $self->_get_droplet_for($event->from_user, $tag) })
     ->then(sub ($droplet) {
       if ($droplet) {
         $Logger->log([ "Created droplet: %s (%s)", $droplet->{id}, $droplet->{name} ]);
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/InABox.pm
```diff
@@ -102,14 +103,19 @@ sub handle_box ($self, $event) {
...
     return $event->error_reply("usage: box [status|create|destroy|shutdown|poweroff|poweron|vpn]");
   }
 
-  $handler->($self, $event, @args)
+  my ($switches, $error) = parse_switches($args);
+  return $event->error_reply("couldn't parse switches: $error") if $error;
+
+  my %switches = map { $_->[0] => ($_->[1] // []) } @$switches;
+
+  $handler->($self, $event, %switches)
```
mmcclimon, line 123: More idiomatic in synergy would be to pass a hashref rather than a list of pairs. (We generally avoid slurpy params, though I realize this was already here..)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -480,8 +479,11 @@ sub _get_ssh_key ($self) {
   );
 }
 
-sub _box_name_for_user ($self, $user) {
-  return sprintf("%s.box.%s", $user->username, $self->box_domain);
+sub _dns_name_for ($self, $user) {
+  return join '.', $user->username, 'box';
+}
+sub _box_name_for ($self, $user) {
```
mmcclimon: (I would stick a blank line before this.)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -258,21 +269,28 @@ sub handle_create ($self, $event, %switches) {
...
 
-  return $self->_get_droplet_for($event->from_user)
+  return $self->_get_droplet_for($event->from_user, $tag)
     ->then(sub ($maybe_droplet) {
-      return Future->fail("You don't have a box.", 'stop-processing')
-        unless $maybe_droplet;
+      unless ($maybe_droplet) {
+        return Future->fail(
+          "That box doesn't exist: " . $self->_box_name_for($event->from_user, $tag),
+          'stop-processing'
```
mmcclimon, line 300: (Man, Synergy has like, 8 different future idioms throughout. We should decide on one and use it, but not as part of this, obviously.)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -308,10 +326,12 @@ sub _handle_power ($self, $event, $action) {
...
 
-  $self->_get_droplet_for($event->from_user)
+  $self->_get_droplet_for($event->from_user, $tag)
     ->then(sub ($maybe_droplet) {
-      return Future->fail("You don't have a box.", 'stop-processing')
-        unless $maybe_droplet;
+      return Future->fail(
+        "That box doesn't exist: " . $self->_box_name_for($event->user, $tag),
+        'stop-processing',
+      ) unless $maybe_droplet;
```
mmcclimon, line 347: I dislike postfix `unless` at the end of multiline statements. I'd stick this in a block, or (more likely) pull out `$name` first and leave it formatted as it was before.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -162,12 +162,13 @@ sub _determine_version_and_tag ($self, $event, %switches) {
   # - if this is the "default" box, which sets the "$username.box" DNS name
   my $default_version = $self->get_user_preference($event->from_user, 'version');
   my ($version, $tag) = @switches{qw(version tag)};
+  my $is_default_box = !($version || $tag);
```
mmcclimon, line 170: I find this condition difficult to read. I think I would probably write:

```
my ($version, $tag) = @switches{qw(version tag)};
$version // = $default;
$tag //= $version
my $is_default = $version eq $default_version && $tag eq $version;
```

Which has the added advantage of setting `$is_default` correctly even if my default is jessie and I say `box create /version jessie`. (Is that an advantage, now that I type it? Maybe, maybe not.)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -270,7 +271,14 @@ sub handle_create ($self, $event, %switches) {
       # else can we do?
       my $ip_address = $self->_ip_address_for_droplet($droplet);
 
-      $self->_update_dns($self->_dns_name_for($event->from_user, $tag), $ip_address);
+      my $f = $self->_update_dns($self->_dns_name_for($event->from_user, $tag), $ip_address);
+      if ($is_default_box) {
```
mmcclimon, line 280: I'm not sure I understand what this block is doing. I think: if we're creating my default box, we'll set up both michael-jessie.box.whateverdomain and michael.box.whateverdomain? If so, I think a comment is probably in order (and probably I would rename the first $f and not reuse that in the wait_all).

lib/Synergy/Reactor/InABox.pm
```diff
@@ -32,15 +32,20 @@ sub listener_specs {
...
+• create: create a new box
+• destroy: destroy a box. if its powered on, you have to shut it down first
 • shutdown: gracefully shut down and power off your box
 • poweroff: forcibly shut down and power off your box (like pulling the power)
 • poweron: start up your box
 • vpn: get an OpenVPN config file to connect to your box
+
+The following preferences exist:
+
+* inabox.version: which version to create by default (default: jessie)
```
mmcclimon, line 47: The first half of these are dependent on config. (Our reactor happens to be called "inabox".) You can get whatever is correct with `->preference_namespace`.

mmcclimon (commented):
Yeah, I think this is mostly fine. I have left some minor fiddly comments.

lib/Synergy/Reactor/InABox.pm
robn: That's right. The commit message said it, but I've added comments and spread the code out a little.

lib/Synergy/Reactor/InABox.pm
robn: (yep)

lib/Synergy/Reactor/InABox.pm
robn: Fair enough, sorted.

lib/Synergy/Reactor/InABox.pm
robn: I went with the block.

lib/Synergy/Reactor/InABox.pm
robn: Yeah, the idea is that if you don't pass args, you're setting up your "default" box, and ultimately, we create the `username.box` DNS name. If you pass args, then you're explicitly saying you know what you're doing, and this isn't the default.

I honestly don't know what the behaviour should be if you provide args that match the default. I sort of don't know exactly what the interface and back-compat bits should be like at all. I thought I'd wait to see how people use it?

lib/Synergy/Reactor/InABox.pm
robn: I will ask you about this when you're on the clock (read: my dumb attempt failed spectacularly and I didn't find a good example to steal from).

robn:
@mmcclimon updated as requested; further commentary above.

lib/Synergy/Reactor/InABox.pm
robn: I will ask here, and then tag you in.

If I do `$self->preference_namespace` in `listener_specs`, I get:

```
Attribute (preference_namespace) does not pass the type constraint because: Validation failed for 'Str' with value undef at reader Synergy::Reactor::InABox::preference_namespace (defined at lib/Synergy/Role/HasPreferences.pm line 31) line 8
[boring trace follows]
```

So its not able to fill in from its default, and I'm not entirely sure what to add to populate whatever its looking for, and there's nothing obvious to steal from.

OR: I'm holding it wrong, and I can't call it from `listener_specs`, or I have to call it through `__PACKAGE__`, or ...? (those weren't obvious either).

Lil help?

mmcclimon: (As discussed in Slack.)

Oh this is some Moose nonsense having to do with the order of role application and object instantiation. If you reran your thing a few times, you'd see it work sometimes and fail that way other times.  I think you can sidestep this entirely by just making `listeners` in ::EasyListening be lazy, which means that we're sure the object is fully initialized by the time we get around to calling `->listener_specs` (which means that `->preference_namespace` won't crash).

robn: Patched synergy-prime as discussed, and this is now working just nicewards.

PR #134, opened 2020-10-30
Clockout to clocks

Davis-A:


Davis-A:
I checked with Nicola (who asked for the feature) if when someone clocks out in a private message with Synergy should it get sent to #clocks.  And she made an excellent point that `@synergy recent clockouts` will show private clockouts anyway.  

I'm thinking because of this we could also scrap the code   
````
if ($event->is_public) {
  $event->reply("See you later!");
} else {
  $event->reply("See you later! Next time, consider clocking \L$w2\E in public!");
}
````
And just always  
````
$event->reply("See you later!")
````

mmcclimon:
Rebased and applied, thanks!

PR #135, opened 2020-11-23
Resolve plumbing

Davis-A:
This will prevent our rerouted incidents from being resolved by `maint end`, `resolve all` and `resolve mine`

I didn't touch `ack all` because we don't have any incidents that route anywhere but the plumbing team.  

Tests also updated.


mmcclimon:
@Davis-A Thanks! I squashed the two test commits into one and applied this.

PR #141, opened 2021-04-18
WIP: set a log prefix inside plugin

rjbs:
Logger prefixes are nice because you know what's logging what, and you don't need to put "logged from this part of the code" in every log line.

They don't work well in async code, though, because dynamic scope doesn't hold up.  Code gets sequenced into futures, using coderefs, which then get executed in some other dynamic scope in the loop.

This branch demonstrates a way to insert the proxy prefix at the moment of logging, by using a closed-over `$self` as a means to get the prefix, but then uses the dynamic $Logger as the target of the logging call.  This means that sequenced coderefs will need to close over `$self`, but that's probably okay.

If we like this, I would move this code into HubComponent and then update existing Channels and Reactors to call the logging methods on $self rather than $Logger.

### DIFF
--- a/lib/Synergy/External/Discord.pm
+++ b/lib/Synergy/External/Discord.pm
@@ -21,6 +21,27 @@ my $JSON = JSON->new;
 
 with 'Synergy::Role::HubComponent';
 
+use Params::Util qw(_HASH0 _ARRAY0);
+BEGIN {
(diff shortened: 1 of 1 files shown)


wolfsage:
Seems good. Not sure about `": "` being included after $self->name, but I also don't see why not

PR #143, opened 2021-06-10
VictorOps reactor: add a maint-on-resolve pref

rjbs:


mmcclimon:
I'm pretty sure I am on record as being against "demaint always resolves" in the first place, and I don't really like this either. What I think would be better is to restore the old behavior of demaint doesn't resolve, unless you pass `/resolve`; then we could have a preference that says "always pass /resolve" (which would be off by default).

The logic here is that: if demaint resolves by default, and you `demaint` and didn't _want_ to resolve, you're SOL. If demaint doesn't resolve by default, and you `demaint`, you can say "oh yeah, that didn't resolve" and just say `resolve all` or whatever. It also stinks, as you say, that there is no way to resolve-on-maint if you set your preference to no.  You'll just need to

IIRC, the reason you wanted demaint to resolve was to resolve everything that happened during maint and 10min before, because demaint is "clear the boards." That's not crazy, but I don't think it should be the default. Maybe what I'd like is: if you say `resolve all` and there was a recent maint, resolve everything since 10m before maint start time, maybe with a thing that says "if you really want to resolve everything, do it again".

(I haven't looked too carefully at the code just yet.)

robn:
As a user, I will say: when I demaint, I usually _don't_ want to resolve, because I only use maint in very specific and extreme situations, and in those I want to demaint as soon as possible but while a large number of alerts are still acked, because I want awareness of new things. Put another way, I demaint when the patient is in a critical-but-stable condition.

(I have strong opinions about the proper use of maint, but I will not re-argue them here).

rjbs:
I really don't care.  What happened is one day I ended maint, resolved things, and missed something.  Andrew ended up getting paged.  I said, "Hm, maybe we should make ending maint resolve things by default?" and then there was some discussion and whoever I talked to said "yeah let's do that."

Then later, somebody else didn't like it and I tried to add an out.

I do not care what the actual behavior here is.  I want people to feel like they have a tool they can make behave sanely, and I don't want to catch grief because I have contrived to provide a tool that doesn't work how they want.

PR #145, opened 2021-06-18
Add DeduplicatesExpandos role, use it

mmcclimon:
This removes the copypasta from the GitLab and Zendesk reactors, and
lets us reuse that in the future for other informational expando things.


mmcclimon:
@rjbs Have a look at this? (Now with the timerless version.)

rjbs (approved):
Looks good!

PR #146, opened 2021-07-08
Fix vo demaint noresolve

Davis-A:
It hasn't been possible to demaint without resolving since the default behaviour was changed.  Was just a minor error in the regex.  Well, it could work if you knew it was demaint /noresolve/ with the trailing slash being required

I've also updated the tests to have a test case for both resolving and not resolving incidents


rjbs (approved):
LGTM!

rjbs:
Applied!

PR #147, opened 2021-07-08
cpanfile: add Safe::ISA

Davis-A:
Required by the inabox reactor


rjbs:
Clearly correct, thanks!

PR #152, opened 2021-08-30
Status Reactor: Report when someone in another time zone is off

wolfsage:
The use of `shift` was a mistake, it should have been `$shift`


mmcclimon (approved):
LGTM.

PR #153, opened 2021-09-02
inabox: remove dns entries on box destroy

Davis-A:


lib/Synergy/Reactor/InABox.pm
```diff
@@ -333,9 +333,12 @@ sub handle_destroy ($self, $event, $switches) {
       }
 
       $droplet = $maybe_droplet;
-      return Future->done($maybe_droplet)
+      return Future->done;
```
robn, line 336: Mmm. I feel like we need a bit of consistency on whether we thread the var through the futures vs use a var scoped over the futures. I prefer the former even though I wrote some of it the other way, but its a mixed bag so I think this is just idle pondering, not actually anything to do!

lib/Synergy/Reactor/InABox.pm
```diff
@@ -591,6 +594,23 @@ sub _region_for_user ($self, $user) {
...
+
+  return $self->_do_request(GET => "$base?per_page=200")
+  ->then(sub ($data) {
+    my @records = grep {$_->{data} eq $ip } $data->{domain_records}->@*;
+    Future->done(@records);
+  })
+  ->then(sub (@records) {
+    return Future->wait_all(
+      map {
+        $self->_do_request(DELETE => "$base/$_->{id}", {})
```
robn, line 608: Pop a little log in here? "Removing DNS entries:" or something?

lib/Synergy/Reactor/InABox.pm
```diff
@@ -591,6 +594,23 @@ sub _region_for_user ($self, $user) {
                            'nyc3';
 }
 
+sub remove_dns_entries ($self, $ip) {
+  my $base = '/domains/' . $self->box_domain . '/records';
+
+  return $self->_do_request(GET => "$base?per_page=200")
```
robn, line 600: I don't think we have 200 records yet, but worth noting that if we get more we'll have to start paging through them.

lib/Synergy/Reactor/InABox.pm
robn: Only works if you actually have all your handlers in the same scope though. If you start to lift things out of a single function, then you lose the scope. And threading them through makes them more function-cally.

I'm really not the guy for this opinion though!

lib/Synergy/Reactor/InABox.pm
Davis-A: Not yet, and this will help keep the number down too so i'm not concerned for now.

Note this line was lifted directly for `_update_dns()` so following the pattern of what's already there.  

Certainly something to keep in mind in the future.

lib/Synergy/Reactor/InABox.pm
Davis-A: I've popped one in `handle_destroy`.  I think that's a reasonable spot because it can log the name of the droplet whose entries are being removed.

lib/Synergy/Reactor/InABox.pm
Davis-A: Referring back to:
>Mmm. I feel like we need a bit of consistency on whether we thread the var through the futures vs use a var scoped over the futures

Before this one variable was done both ways.  Now it's scoped consistently in the scope of the handler.  

If we want to clean it up to a standard everywhere, that's fine but I think that's outside the scope of this PR.  For now I think this is fine

robn:
No showstoppers, so I think we're good to go!

PR #154, opened 2021-09-03
Bort backports

mmcclimon:
I want to rebase bortergy onto synergy to get a year's worth of minor improvements, and realized that there were some things that would be good to backport into synergy prime.

### DIFF
--- a/lib/Synergy/Util.pm
+++ b/lib/Synergy/Util.pm
@@ -386,12 +386,12 @@ my %Trans = (
   fraktur => _wonky_style('fraktur'),
   sans    => _wonky_style('ss'),
   double  => _wonky_style('double'),
-
-  zalgo   => sub ($s) {
-    require Acme::Zalgo;
-    Acme::Zalgo::zalgo($s, 0, 2, 0, 0, 0, 2);
-  },
 );
+
+eval "require Acme::Zalgo";
+if (!$@) {
+  $Trans{zalgo} = sub ($s) { Acme::Zalgo::zalgo($s, 0, 2, 0, 0, 0, 2); };
+}
 
 sub _wonky_style ($style) {
   my $i = 0;
--- a/lib/Synergy/Util.pm
+++ b/lib/Synergy/Util.pm
@@ -12,6 +12,7 @@ use DateTime::Format::Natural;
 use JSON::MaybeXS;
 use List::Util qw(first any);
 use Path::Tiny ();
+use Synergy::Logger '$Logger';
 use Time::Duration::Parse;
 use Time::Duration;
 use TOML;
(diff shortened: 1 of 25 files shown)


lib/Synergy/Reactor/VictorOps.pm
```diff
@@ -437,13 +437,11 @@ sub handle_maint_query ($self, $event) {
...
       }
 
-      state $ago_formatter = DateTimeX::Format::Ago->new(language => 'en');
-
       # the way we use VO there's probably not more than one, but at least this
       # way if there are we won't just drop the rest on the floor -- robn, 2019-08-16
       my $maint_text = join ', ', map {
         my $start = DateTime->from_epoch(epoch => int($_->{startedAt} / 1000));
-        my $ago = $ago_formatter->format_datetime($start);
+        my $ago = ago(time - $start->epoch);
```
wolfsage, line 444: Why do the above changes to use `ago(...)` just swap out `ago()` for `$ago_formatter->format_datetime()` but this one first subtracts the argument from `time()`?

lib/Synergy/Util.pm
```diff
@@ -384,10 +383,13 @@ my %Trans = (
...
   sans    => _wonky_style('ss'),
   double  => _wonky_style('double'),
-
-  zalgo   => sub ($s) { Acme::Zalgo::zalgo($s, 0, 2, 0, 0, 0, 2); },
 );
 
+eval "require Acme::Zalgo";
+if (!$@) {
+  $Trans{zalgo} = sub ($s) { Acme::Zalgo::zalgo($s, 0, 2, 0, 0, 0, 2); };
+}
```
wolfsage, line 391: I would do something if zalgo doesn't work - warn, log, or fill in a sub that just says "sorry ask someone install Acme::Zalgo"

lib/Synergy/Reactor/VictorOps.pm
mmcclimon: Good catch, thanks! This one was right, the others were wrong. That means the GitLab reactor has been wrong in bort for a year at least, which we might have noticed if bort ever used the GitLab reactor.

I have amended this commit.

lib/Synergy/Util.pm
mmcclimon: Fair enough. I added a commit to address this, which also lazily loads all the alphabets. (The reason I didn't add this before is because I didn't want a log line every time Bort starts up that Acme::Zalgo isn't loaded, because it doesn't even use the alphabet stuff.)

mmcclimon:
@wolfsage I amended one commit and added a new on on top. Thoughts?

lib/Synergy/Util.pm
```diff
@@ -308,85 +308,96 @@ my %MORSE_FOR = do {
...
+    fraktur => _wonky_style('fraktur'),
+    sans    => _wonky_style('ss'),
+    double  => _wonky_style('double'),
+  );
 
-  zalgo   => sub ($s) { Acme::Zalgo::zalgo($s, 0, 2, 0, 0, 0, 2); },
-);
+  eval "require Acme::Zalgo";
+  if ($@) {
+    $Logger->log("ignoring Zalgo alphabet because Acme::Zalgo isn't installed");
```
wolfsage, line 396: $@ doesn't necessarily mean it isn't installed, but good enough I suppose!

wolfsage (approved):

PR #156, opened 2021-09-23
pd: announce @oncall changes

Davis-A:
With VictorOps the slack integration announced when there is an oncall
change.

Pager Duty does it silently.  Synergy can pick up the slack (pun intended).

Anyway, not sure how many people found the announcement useful.  I know I did, and speaking to Butch he does too.

### DIFF
--- a/lib/Synergy/Reactor/PagerDuty.pm
+++ b/lib/Synergy/Reactor/PagerDuty.pm
@@ -820,47 +820,36 @@ sub _check_at_oncall ($self) {
         $self->_set_oncall_list(\@new);
       });
 
-      $self->_announce_oncall_change({were => \@have, now => \@new});
+      $self->_announce_oncall_change(\@have, \@new)
         if $self->oncall_change_announce_address;
 
       return $f;
     })->retain;
 }
 
-sub _announce_oncall_change($self, $args) {
-  my %were = map {
-    $self->username_from_pd($_) ? $self->username_from_pd($_) : $_ => 1
-  } $args->{were}->@*;
-  
-  my %are = map {
-    $self->username_from_pd($_) ? $self->username_from_pd($_) : $_ => 1
-  } $args->{now}->@*;
+sub _announce_oncall_change ($self, $old, $new) {
+  my %before = map {; ($self->username_from_pd($_) // $_) => 1 } $old->@*;
+  my %after = map {; ($self->username_from_pd($_) // $_) => 1 } $new->@*;
 
-  my @leaving = grep {
-    ! $are{$_};
-    } keys %were;
-
-  my @joining = grep {
-  ! $were{$_};
-  } keys %are;
+  my @leaving = grep { ! $after{$_} } keys %before;
+  my @joining = grep { ! $before{$_} } keys %after;
 
   my @lines;
 
   if (@leaving) {
-    my $verb = @leaving > 1
-             ? 'have'
-             : 'has';
+    my $verb = @leaving > 1 ? 'have' : 'has';
     my $removed = join ', ', sort @leaving;
     push @lines, "$removed $verb been removed from the oncall group";
   }
 
   if (@joining) {
-    my $verb = @joining > 1
-             ? 'have'
-             : 'has';
+    my $verb = @joining > 1 ? 'have' : 'has';
     my $added = join ', ', sort @joining;
     push @lines, "$added $verb been added to the oncall group";
   }
+
+  my $oncall = join ', ', sort keys %after;
+  push @lines, "Now oncall: $oncall";
 
   my $message = join "\n", @lines;
   $self->oncall_channel->send_message(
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -815,10 +820,55 @@ sub _check_at_oncall ($self) {
...
+      $self->_announce_oncall_change({were => \@have, now => \@new});
+        if $self->oncall_change_announce_address;
+
       return $f;
     })->retain;
 }
 
+sub _announce_oncall_change($self, $args) {
+  my %were = map {
+    $self->username_from_pd($_) ? $self->username_from_pd($_) : $_ => 1
```
mmcclimon, line 832: Here (and below): `username_from_pd` returns a username or undef, so you could write this without the ternary:

```perl
my %old = map {; ($self->username_from_pd($_) // $_) => 1 } $args->{were}->@*;
```

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -815,10 +820,55 @@ sub _check_at_oncall ($self) {
...
+  } $args->{were}->@*;
+  
+  my %are = map {
```
mmcclimon, line 835: (nit) I find the names `were` and `are` to be hard to understand; I'd call them "old" and "new".

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -815,10 +820,55 @@ sub _check_at_oncall ($self) {
...
+    $self->username_from_pd($_) ? $self->username_from_pd($_) : $_ => 1
+  } $args->{now}->@*;
+
+  my @leaving = grep {
```
mmcclimon, line 839: Style: this should  go on one line, and you could write the whole opening of this sub much more concisely!

```perl
  my %old = map {; ($self->username_from_pd($_) // $_) => 1 } $args->{were}->@*;
  my %new = map {; ($self->username_from_pd($_) // $_) => 1 } $args->{are}->@*;

  my @leaving = grep { ! $new{$_} } keys %old;
  my @joining = grep { ! $old{$_} } keys %new;
```

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -815,10 +820,55 @@ sub _check_at_oncall ($self) {
...
+    } keys %were;
+
+  my @joining = grep {
+  ! $were{$_};
+  } keys %are;
+
+  my @lines;
+
+  if (@leaving) {
+    my $verb = @leaving > 1
```
mmcclimon, line 850: Style: short ternaries go on a single line.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -815,10 +820,55 @@ sub _check_at_oncall ($self) {
...
+
+  if (@joining) {
+    my $verb = @joining > 1
+             ? 'have'
+             : 'has';
+    my $added = join ', ', sort @joining;
+    push @lines, "$added $verb been added to the oncall group";
+  }
+
+  my $message = join "\n", @lines;
```
mmcclimon, line 865: I wonder if it would be useful to include a line at the end that says: `"now oncall: " . join q{, } sort @$new`. (I often find this lacking in the VO integration, because it's like "ok, andrew and marc are off so I can assume rob is on now I guess".

mmcclimon (commented):
I have suggested one actual change and several style things, but in general this is fine!

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -815,10 +820,55 @@ sub _check_at_oncall ($self) {
...
       });
 
+      $self->_announce_oncall_change({were => \@have, now => \@new});
+        if $self->oncall_change_announce_address;
+
       return $f;
     })->retain;
 }
 
+sub _announce_oncall_change($self, $args) {
```
mmcclimon, line 830: Oh also: you're only ever going to call this in one place, so I'd just have this take two positional args rather than named params (so `sub _announce_oncall_change ($self, $old, $new) { ... }`). At any rate, you need a space before the opening paren!

lib/Synergy/Reactor/PagerDuty.pm
Davis-A: I really dislike `new` because it implies that they're newly put oncall.  But in a case like i"m oncall, then you come on call with me `new` implies we're both newly oncall.  I wouldn't be new, i was already oncall.

I'm not new, i'm oncall.  How about `old` and `now` ?

mmcclimon: `old` and `now` is worse than `were` and `are` because _every_ time I'm going to mistype it as `old` and `new`! It's not "newly oncall", it's "the old list of people we had oncall" and "the new list of people we have oncall". But my feelings aren't _that_ strong.

Davis-A: `before` and `after` make perfect sense!

lib/Synergy/Reactor/PagerDuty.pm
Davis-A: done!

PR #159, opened 2021-11-06
inabox: move default datacentre for AU sfo2 --> sfo3

Davis-A:
There are some droplet types available in sfo3 that Butch wants for the
automated testing work.  It doesn't make sense to have snapshots at both
sfo2 and sfo3 long term so update the default so we can remove the snapshots from sfo2.

There will be some users with a preference set for sfo2.  Would it be best to "send an email" to get everyone with it manually set to update it, or just update it from underneath them (and if so, how?)


robn:
Obviously correct.

For changing them, I would probably change it out from under them and also send them mail. Better than no-worky at all.

robn:
Approved or whatever.

PR #161, opened 2021-12-22
Linear: Add 'Lagenda' cmd, takes no args, user, team, or user@team

wolfsage:
This does *not* honor the default team of the user, so that asking
for a user's agenda finds their items across all teams.


rjbs:
Great, thanks!

PR #162, opened 2021-12-27
Add Synergy::Reactor::LinearNotification - handles triage tasks

wolfsage:
That is, any task created with the configured label will send
a message to the configured slack channel about it

### DIFF
--- a/lib/Synergy/Reactor/LinearNotification.pm
+++ b/lib/Synergy/Reactor/LinearNotification.pm
@@ -1,0 +1,184 @@
+use v5.24.0;
+use warnings;
+use utf8;
+package Synergy::Reactor::LinearNotification;
+
+use Moose;
+with 'Synergy::Role::Reactor::EasyListening';
+with 'Synergy::Role::HTTPEndpoint';
+
+use experimental qw(signatures postderef);
+use namespace::clean;
+use Synergy::Logger '$Logger';
+
+use JSON::MaybeXS qw(encode_json decode_json);
+use Try::Tiny;
+use Synergy::Reactor::Linear;
+
+my $ESCALATION_EMOJI = "\N{HELMET WITH WHITE CROSS}";
+
+has '+http_path' => (
+  default => '/linear-notification',
+);
+
+has escalation_label_name => (
+  is => 'ro',
+  isa => 'Str',
+  default => 'support blocker',
+);
+
+# Synergy channel, like 'slack' or w/e it's called in config
+has escalation_channel_name => (
+  is  => 'ro',
+  isa => 'Str',
+);
+
+# The slack address of the channel to send notificaitons to
+has escalation_address => (
+  is  => 'ro',
+  isa => 'Str',
+);
+
+sub _rototron ($self) {
+  # TODO: indirection through rototron_reactor name on object
+  return unless my $roto_reactor = $self->hub->reactor_named('rototron');
+  return $roto_reactor->rototron;
+}
+
+my %allowed = (
+  '35.231.147.226' => 1,
+  '35.243.134.228' => 1,
+);
+
+has confirm_remote_ips => (
+  is => 'ro',
+  isa => 'Bool',
+  default => 1,
+);
+
+has linear_token => (
+  is => 'ro',
+  required => 1,
+);
+
+has linear => (
+  is => 'ro',
+  isa => 'Linear::Client',
+  default => sub ($self) {
+    Linear::Client->new({
+      auth_token      => $self->linear_token,
+      debug_flogger   => $Logger,
+
+      helper => Synergy::Reactor::Linear::LinearHelper->new_for_reactor($self),
+    });
+  },
+  lazy => 1,
+);
+
+has name => (
+  is => 'ro',
+  default => 'lin',
+);
+
+sub http_app ($self, $env) {
+  my $req = Plack::Request->new($env);
+
+  if ($self->confirm_remote_ips) {
+    unless ($allowed{$req->address}) {
+      warn "ADDRESS " . $req->address . " not allowed\n";
+      return [ "200", [], [ '{"go":"away"}' ] ];
+    }
+  }
+
+  my $err;
+
+  my $payload = try {
+    decode_json( $req->raw_body );
+  } catch {
+    $err = $_;
+  };
+
+  if ($err) {
+    warn "Failed to parse json: $err\n";
+
+    return [ "200", [], [ '{"bad":"json"}' ] ];
+  }
+
+  if ($self->escalation_channel_name && $self->escalation_address) {
+    if (my $channel = $self->hub->channel_named($self->escalation_channel_name)) {
+
+      my $was_create = $payload->{type} eq 'Issue'
+                    && $payload->{action} eq 'create'
+                    && grep {
+                         lc $_->{name} eq lc $self->escalation_label_name
+                       } $payload->{data}->{labels}->@*;
+
+      my $was_update;
+
+      if (   ! $was_create
+          && $payload->{type} eq 'Issue'
+          && $payload->{action} eq 'update'
+      ) {
+        # Did we add the escalation label to an existing task?
+        my ($label) = grep {
+          lc $_->{name} eq lc $self->escalation_label_name
+        } $payload->{data}->{labels}->@*;
+
+        if ($label) {
+          my $id = $label->{id};
+          if (   $payload->{updatedFrom}{labelIds}
+              && ! grep { $_ eq $id } $payload->{updatedFrom}{labelIds}->@*
+          ) {
+            $was_update = 1;
+          }
+        }
+      }
+
+      if ($was_create || $was_update) {
+        $self->linear->users->then(sub ($users) {
+          my %by_id = map { $users->{$_}->{id} => $users->{$_} } keys %$users;
+
+          return Future->done($by_id{$payload->{data}->{creatorId}}->{displayName} // 'unknown');
+        })->then(sub ($who) {
+          my $desc = $was_create ? 'New task created for' : 'Existing task moved to';
+
+          my $text = sprintf
+            "$ESCALATION_EMOJI %s escalation by %s in zendesk: %s (%s)",
+            $desc,
+            $who,
+            $payload->{data}{title},
+            $payload->{url};
+
+          if (my $rototron = $self->_rototron) {
+            my $roto_reactor = $self->hub->reactor_named('rototron');
+
+            for my $officer ($roto_reactor->current_triage_officers) {
+              $Logger->log(["notifying %s of new escalation task", $officer->username ]);
+              $channel->send_message_to_user($officer, $text);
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+);
+
+has name => (
+  is => 'ro',
+  default => 'lin',
+);
+
+sub http_app ($self, $env) {
+  my $req = Plack::Request->new($env);
+#  unless ($allowed{$req->address}) {
```
wolfsage, line 80: If we actually get the real remote ip we need to check it here (basically just uncomment this)

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+  is => 'ro',
+  required => 1,
+);
+
+has _linear_shared_cache => (
+  is => 'ro',
+  default => sub {  {}  },
+);
+
+
```
rjbs, line 57: recently earned your doctorate, eh?

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
+use v5.24.0;
+use warnings;
+use utf8;
+package Synergy::Reactor::LinearNotification;
+
+use Moose;
+with 'Synergy::Role::Reactor::EasyListening';
+with 'Synergy::Role::HTTPEndpoint';
```
rjbs, line 8: Probably makes no difference, but it's almost always better to pass multiple arguments to `with` rather than to call it twice.  That way it will detect composition conflicts.

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+
+use experimental qw(signatures postderef);
+use namespace::clean;
+use Synergy::Logger '$Logger';
+
+use JSON 2 qw(encode_json decode_json);
```
rjbs, line 14: As of 5m ago, Synergy always uses JSON::MaybeXS everywhere.  Please amend.
```suggestion
use JSON::MaybeXS qw(encode_json decode_json);
```

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+use Try::Tiny;
+use Synergy::Reactor::Linear;
+
+my $TRIAGE_EMOJI = "\N{HELMET WITH WHITE CROSS}";
+
+has '+http_path' => (
+  default => '/linear_notification/',
```
rjbs, line 21:
```suggestion
  default => '/linear-notification/',
```

LOW LINE is bad in general, but especially in URLs.

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+);
+
+has triage_label_name => (
+  is => 'ro',
+  isa => 'Str',
+  default => 'support blocker',
+);
```
rjbs, line 28: We absolutely must stop using "triage" to mean the old thing!  Yes, this means renaming hunks of code, updating documentation, and so on.  The alternative is explaining this confusing situation to all new hires forever!

What if, instead, configuration was like this:
```
label_subscriptions => {
  'support blocker' => [ [ slack => 'C1234' ], [ twilio => '16105551212' ] ],
  'printer fire' => [ ... ],
}
```

Then multiple labels could be noticed, sending to multiple targets.  Even if we start off only wanting one!

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+# The slack address of the channel to send notificaitons to
+has triage_address => (
+  is  => 'ro',
+  isa => 'Str',
+);
+
+my %allowed = (
+  '35.231.147.226' => 1,
+  '35.243.134.228' => 1,
+);
```
rjbs, line 45: This is fine, mayyyybe, but do they guarantee this won't change?  Seems unlikely, they're just hosted in the Google cloud.  Instead:
1. I'd hope their webhooks would include some verifiable data!
2. if not, we could add a mandatory path element that is a GUID and can't be guessed, then register that

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+
+
+has linear => (
+  is => 'ro',
+  isa => 'Linear::Client',
+  default => sub ($self) {
+    Linear::Client->new({
+      auth_token      => $self->linear_token,
+      _cache_guts     => $self->_linear_shared_cache,
```
rjbs, line 64: I doubt you need a shared cache, since you're only creating one Linear::Client for the life of the reactor.

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+      helper => Synergy::Reactor::Linear::LinearHelper->new_for_reactor($self),
+    });
+  },
+  lazy => 1,
+);
+
+has name => (
+  is => 'ro',
+  default => 'lin',
+);
```
rjbs, line 76: Is this ever used?

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+  };
+
+  if ($err) {
+    warn "Failed to parse json: $err\n";
+
+    return [ "200", [], [ '{"bad":"json"}' ] ];
+  }
+
+#  use Data::Dumper;
+#  warn Dumper $payload;
```
rjbs, line 100: noise to be deleted

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+
+  if ($self->triage_channel_name && $self->triage_address) {
+    if (my $channel = $self->hub->channel_named($self->triage_channel_name)) {
+      if (
+          $payload->{type} eq 'Issue'
+       && $payload->{action} eq 'create'
+       && grep { lc $_->{name} eq $self->triage_label_name } $payload->{data}->{labels}->@*
```
rjbs, line 107: Nothing required that the label name was lc, so you should lc it here.  But since I already said "maybe this whole bit should be genericized" I can withhold further comments…

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+      ) {
+        $self->linear->users->then(sub ($users) {
+          my %by_id = map { $users->{$_}->{id} => $users->{$_} } keys %$users;
+
+          return Future->done($by_id{$payload->{data}->{creatorId}}->{displayName} // 'unknown');
+        })->then(sub ($who) {
+          my $text = sprintf
+            "$TRIAGE_EMOJI New task created for triage by %s in %s: %s (%s)",
```
rjbs, line 115: in a "subscribe to multiple labels" world, we'd replace "for triage" with "with label X"

if we don't enter that world, we still need to expunge "triage"

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+            $who,
+            "zendesk",
```
rjbs, line 117: This isn't really a constant, right?

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+            $payload->{data}{title},
+            $payload->{url};
+
+          return $channel->send_message($self->triage_address, $text);
+        })->catch(sub (@) {
```
rjbs, line 122: You could just leave off the signature, you know!

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+
+sub listener_specs {
+  return {
+    name      => 'linear_notification',
+    method    => 'linear_notification',
+    predicate => sub ($, $e) { 0 },
+  };
+}
+
+sub start ($self) {}
```
rjbs, line 142: Surprised we need this!

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -0,0 +1,144 @@
...
+  return [ "200", [], [ '{"o":"k"}' ] ];
+}
+
+sub listener_specs {
+  return {
+    name      => 'linear_notification',
+    method    => 'linear_notification',
+    predicate => sub ($, $e) { 0 },
+  };
+}
```
rjbs, line 140: This doesn't make sense to me.  If it never matches, why are we registering a listener at all?  Also, I don't think there's a method with that name.  I think this whole sub is a mistake?

rjbs (changes requested):

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: Yeah, like half of our reactors do it this way already. Could "fix" them all at some point...

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: Pff. Fine

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: This is how they document "how to know it's from us!" Sadly there's no other ... thing (a shared secret, etc). Certainly adding a guid makes it less discoverable, but we should really do both.

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: Yeah, I mean, a lot of this was copy pasta and I didn't bother renaming everything in the interest of getting it working. But I can, sure.

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: Will fix

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: Fair. I had had a signature which broke shit so I backed it off to this but will just remove

lib/Synergy/Reactor/LinearNotification.pm
wolfsage: We have to be a listener to be a reactor I think? And in the future the reactor *might* handle some things, so for now this makes it all work without having to think/try too hard. I can document it if that's good enough

wolfsage:
Updated to cover most things - also will now notify if someone adds the label to an existing ticket

rjbs (approved):

PR #163, opened 2021-12-28
Linear Reactor: fix label lookup for ptn blocked

RashaMou:
The linear client now looks up label ids based on the team.


rjbs:
cherry picked, thanks!

PR #165, opened 2022-01-01
replace HTTPServer path-to-app map with Plack URLMap

rjbs:


lib/Synergy/HTTPServer.pm
```diff
@@ -53,41 +53,51 @@ has http_server => (
...
+  my @known = $self->registered_paths;
+
+  my ($is_prefix_of) = grep {; index($_, $path) == 0 } @known;
+  confess "refusing to register $path because it is a prefix of $is_prefix_of"
+    if $is_prefix_of;
+
+  my ($is_suffix_of) = grep {; index($path, $_) == 0 } @known;
+  confess "refusing to register $path because it is a suffix of $is_suffix_of"
+    if $is_suffix_of;
+
```
wolfsage, line 90: This prevention seems all wrong to me but I think we can live with it until it's a problem.

That is, it prevents registering /foo and /foot which are distinct paths. URL map will handle /foo and /foo/*, and /foot and /foot/* separately, but not /foobar just fine.

wolfsage (approved):

PR #166, opened 2022-01-02
PotentialReaction: a replacement for Listeners

rjbs:
This will be part of replacing Listeners.  A reactor can reply a number of potential reactions, as long as there is no more than a single exclusive one.  These are then treated like matching listeners.  This moves the matching process into the reactor instead of the listeners.  In reality, all our current reactors use EasyListening, so the logic will live there for now.


wolfsage:
Seems good

wolfsage (approved):

rjbs:
(got rebased and merged)

PR #167, opened 2022-01-02
CommandPost: a new framework for coding reactions

rjbs:
CommandPost: a new framework for coding reactions

In previous versions of Synergy, all reaction was governed by Listener objects, which set the form for how reaction could occur.  This was replaced by the "potential_reactions_to" method, which more readily allowed reactors to choose their own means by which to react to events.

CommandPost is proof of concept of a new way to build reactors.  It puts more things in one place.  It simplifies the common cases of writing named commands and mention-listeners.  It helps avoid duplication, for example repeating a regex in the reaction body as well as the EasyListening-style predicate sub.

Instead of predicates, CommandPost provides "matchers", which return a value that can then be fed to the reaction method, allowing for single-pass parsing (for example).

In addition to providing an implementation of "potential_reactions_to", CommandPost provides four declarative helpers:

* command NAME => {ARG} => METHOD

This declares a command: a reaction that must be targeted and exclusive, and which responds when a message starts with the command name.

* listener NAME => METHOD

This declares a listener: a non-exclusive reaction that gets called on every event.

* reaction NAME => {ARG} => METHOD

This declares a flexible reaction that may be target and may be exclusive.  It's similar to the listener declarations in EasyListening.

* help TITLE => TEXT

This adds a help entry to the reactor, in case you want to add help topics that don't correspond to a command.

### DIFF
--- a/lib/Synergy/CommandPost.pm
+++ b/lib/Synergy/CommandPost.pm
@@ -58,6 +58,13 @@ package Synergy::CommandPost::Object {
       Sub::Install::install_sub({
         as    => "add_$thing",
         code  => sub ($self, $name, $arg, $method) {
+          # Look, there's not reason you can't actually have two listeners,
+          # named Zorch and zorch.  But there's no reason you really ought to
+          # do that either.  So don't.  I'll be proscribing any
+          # case-insensitive conflict for the same of commands, where it _does_
+          # matter. -- rjbs, 2022-01-14
+          $name = lc $name;
+
           Carp::confess("already have a $thing named $name")
             if $self->$check($name);
 
@@ -95,6 +102,7 @@ package Synergy::CommandPost::Object {
     ### First, is there a command to match the first word of the message?
     if ($targeted) {
       my ($first, $rest) = split /\s+/, $event->text, 2;
+      $first = lc $first;
 
       if (my $command = $self->_command_named($first)) {
         my $method = $command->{method};
--- a/lib/Synergy/CommandPost.pm
+++ b/lib/Synergy/CommandPost.pm
@@ -1,0 +1,191 @@
+use v5.24.0;
+use warnings;
+package Synergy::CommandPost;
+
+use experimental 'signatures';
+
+use Synergy::PotentialReaction;
+
+use Sub::Exporter -setup => {
+  groups => { default => \'_generate_command_system' },
+};
+
+package Synergy::CommandPost::Object {
+  use Moose;
+
+  use experimental 'signatures';
+
+  has is_final => (is => 'rw');
+
+  has commands => (
+    isa => 'HashRef',
+    init_arg  => undef,
+    default   => sub {  {}  },
+    traits    => [ 'Hash' ],
+    handles   => {
+      _command_named    => 'get',
+      _register_command => 'set',
+    },
+  );
+
+  has listeners => (
+    isa => 'HashRef',
+    init_arg  => undef,
+    default   => sub {  {}  },
+    traits    => [ 'Hash' ],
+    handles   => {
+      _listener_kv    => 'kv',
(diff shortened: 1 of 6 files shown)


rjbs:
I had planned to rework this, replacing the current "install a bunch of closures" mechanism with something using MXRP.  I had a go at that, and I don't think it will work.  The subroutines are installed too late to be called as declarators, and there's just not a good way (that I know) to make one thing set up helper functions and a role at once.  I'll look into it more, but really we can rework the guts in the future without worrying about them now.  I think the basics are right, and the helpers are right.  This will let us clean up a lot of code to make it easier to read and maintain, so I think we should go ahead more or less with this.

lib/Synergy/CommandPost.pm
```diff
@@ -0,0 +1,191 @@
...
+      return;
+    },
+    help => sub ($name, $text) {
+      $object->add_help($name, {}, $text);
+      return;
+    },
+    listener => sub ($name, $code) {
+      $object->add_listener($name, {}, $code);
+
+      return;
```
mmcclimon, line 175: (nit) some of these have empty lines before the return and some don't.

lib/Synergy/CommandPost.pm
```diff
@@ -0,0 +1,191 @@
...
+
+    return @reactions;
+  }
+
+  no Moose;
+}
+
+sub _generate_command_system ($class, $, $arg, @) {
+  my $object = Synergy::CommandPost::Object->new;
+  return {
```
mmcclimon, line 152: I wonder if it would be useful to export something to give you a handle to the CommandPost object, something like:

```perl
command_post => sub ($reactor) { $object }
```

I admit that right this second I don't know what I'd do with such a thing, but I can imagine, in some reactor, wanting to say `$self->command_post->_command_named` or...something.

lib/Synergy/CommandPost.pm
```diff
@@ -0,0 +1,191 @@
...
+
+  sub potential_reactions_to ($self, $reactor, $event) {
+    my @reactions;
+    my $targeted = $event->was_targeted;
+
+    ### First, is there a command to match the first word of the message?
+    my ($first, $rest) = split /\s+/, $event->text, 2;
+
+    if ($targeted) {
+      if (my $command = $self->_command_named($first)) {
```
mmcclimon, line 99: should you `lc $first` here? And probably also disallow commands that casefold to the same thing.

lib/Synergy/CommandPost.pm
```diff
@@ -0,0 +1,191 @@
...
+        reactor => $reactor,
+        name    => "listener-$listener_pair->[0]",
+        is_exclusive  => 0,
+        event_handler => sub { $reactor->$method($event) },
+      });
+    }
+
+    ### Finally, reactions are the last-resort flexible option.
+    my @reaction_kv = $self->_reaction_kv;
+    @reaction_kv = grep {; $_->[1]{targeted} } @reaction_kv if ! $targeted;
```
mmcclimon, line 128: This seems backward to me. It our event wasn't targeted, we look only for targeted reaction handlers?

mmcclimon (commented):
I think this is mostly right; I had a few questions but I imagine we'll fix any bugs as we find them!

lib/Synergy/CommandPost.pm
rjbs: For now: YAGNI

lib/Synergy/CommandPost.pm
rjbs: good call, fixed

lib/Synergy/CommandPost.pm
rjbs: Whoops!

PR #168, opened 2022-01-03
Linear: make synergy respond to issue identifiers

RashaMou:
In slack, any mention of an issue id expands to show information about
that issue and a link

### DIFF
--- a/lib/Synergy/Reactor/Linear.pm
+++ b/lib/Synergy/Reactor/Linear.pm
@@ -16,8 +16,21 @@ use Synergy::Logger '$Logger';
 
 use utf8;
 
-sub listener_specs {
+sub listener_specs ($self){
+
+  my $issue_id = $self->_issue_regex; 
+  
   return (
+    {
+      name      => 'mentions',
+      method    => 'handle_mentions',
+      exclusive => 1,
+      targeted  => 1,
+      predicate => sub ($, $e) { 
+        return 1 if $e->text =~ /$issue_id/;  
+        return;
+      } 
+    },
     {
       name      => 'list_teams',
       method    => 'handle_list_teams',
@@ -181,6 +194,14 @@ has _linear_shared_cache => (
   default => sub {  {}  },
 );
 
+has _issue_regex => (
+  is => 'ro',
+  lazy => 1,
+  default => sub {
+    return qr/(?:^|\W)(plumb-[0-9]*)/i; # we're matching only plumbing here for now
+  }
+);
+
 sub _with_linear_client ($self, $event, $code) {
   my $user = $event->from_user;
 
@@ -204,6 +225,12 @@ sub _with_linear_client ($self, $event, $code) {
   });
 
   return $code->($linear);
+}
+
+sub handle_mentions ($self, $event) {
+  $event->mark_handled if $event->was_targeted;
+  my $issue_id = $event->text =~ $self->_issue_regex;
+  $event->reply("https://linear.app/fastmail/issue/$1/...");
 }
 
 sub handle_list_teams ($self, $event) {


RashaMou:
@rjbs this is almost ready, but we need to merge the [issue lookup PR](https://github.com/fastmail/Linearsonist/pull/23) in Linearsonist to complete it.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -16,8 +16,20 @@ use Synergy::Logger '$Logger';
...
 use utf8;
 
-sub listener_specs {
+sub listener_specs ($self){
+
   return (
+    {
+      name      => 'mentions',
+      method    => 'handle_mentions',
+      exclusive => 1,
```
rjbs, line 25: Can't be targeted or it'll only fire when someone says "Synergy: BAK-123".  Can't be exclusive or you can't say "++ fix the second problem seen in BAK-321"

lib/Synergy/Reactor/Linear.pm
```diff
@@ -233,6 +245,24 @@ sub _with_linear_client ($self, $event, $code) {
...
+  sprintf "<%s|%s>\N{THIN SPACE}",
+    "https://linear.app/fastmail/issue/$issue->{team}-$issue->{number}/...",
+    $issue->{team}."-".$issue->{number};
+}
+
+sub handle_mentions ($self, $event) {
+  $event->mark_handled if $event->was_targeted;
+  my $issue_id = $event->text;
+  $self->_with_linear_client($event, sub ($linear) {
+      $linear->fetch_issue($issue_id)->then(sub ($issue) {
```
rjbs, line 258: this is 4 sp instead of 2 sp indents

lib/Synergy/Reactor/Linear.pm
```diff
@@ -233,6 +245,24 @@ sub _with_linear_client ($self, $event, $code) {
   return $code->($linear);
 }
 
+sub _slack_item_link ($self, $issue) {
+  sprintf "<%s|%s>\N{THIN SPACE}",
+    "https://linear.app/fastmail/issue/$issue->{team}-$issue->{number}/...",
+    $issue->{team}."-".$issue->{number};
```
rjbs, line 251: You could use $issue->{identifier} instead, which is already in the form XYX-123.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -233,6 +245,24 @@ sub _with_linear_client ($self, $event, $code) {
...
+}
+
+sub handle_mentions ($self, $event) {
+  $event->mark_handled if $event->was_targeted;
```
rjbs, line 255: (Oh, I should make this a generic facility of CommandPost!  For now, keep this line…)

lib/Synergy/Reactor/Linear.pm
rjbs, line 258: You don't have any handling for bogus issues.  What if I say "This will be fixed in BAK-1232134123"?  That issue won't exist and I can't tell what will happen here!

rjbs: With changes suggested in L-C, $issue would now be undef, so you would want to:
* if targeted, reply_error "no such issue, dawg"
* if not targeted, just return done

rjbs (changes requested):

lib/Synergy/Reactor/Linear.pm
RashaMou: I made this change, but could use double checking to make sure I understood correctly.

rjbs:
I will rebase this on the CommandPost changes and apply, thanks!

lib/Synergy/Reactor/Linear.pm
```diff
@@ -233,6 +243,31 @@ sub _with_linear_client ($self, $event, $code) {
...
+  my $issue_id = $event->text;
+  $self->_with_linear_client($event, sub ($linear) {
+    $linear->fetch_issue($issue_id)->then(sub ($issue) {
+      if (defined $issue) {
+        my $title = $issue->{title};
+        my $link = $self->_slack_item_link($issue);
+        return $event->reply("$link: $title");
```
rjbs, line 260: The Slack formatted link has to be given as an "alternative" display, because you don't want to SMS somebody something like `<http:/...foo../|Foo 123>`.  You want:
```
$event->reply($plaintext, { alt_type => $value, alt_type_2 => $value_2, ... });
```

rjbs (changes requested):
I have rebased this, as promised, on the commandpost-extra branch.  I'll walk you through my changes tomororw.

PR #169, opened 2022-01-07
TimeClock: record last report time per-user

mmcclimon:
Sometimes, we get into some state where a newly added or deleted user
causes the time-clock reactor to crash midway through reports. This
means that all the users _before_ the crash often get a skrillion
morning/evening reports, because we boot up, start sending reports,
then crash before setting our "last report" time, over and over again.

That stinks for the users who get a million reports, so instead we'll
store last report time by user. This is slower, because we save our
state to the sqlite database after every user, but they're reports and
it doesn't matter if they're slow!

### DIFF
--- a/lib/Synergy/Reactor/TimeClock.pm
+++ b/lib/Synergy/Reactor/TimeClock.pm
@@ -88,6 +88,7 @@ has clock_out_channel => (
 sub state ($self) {
   return {
     last_report_time => $self->last_report_time,
+    user_last_report_times => $self->user_last_report_times,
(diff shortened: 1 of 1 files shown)


rjbs:
I think this is all good.  I have a niggling suspicion that we could eliminate the global state pretty easily, but it doesn't seem worth the effort.  👍🏽 Everybody who has gotten 12039 reports in an hour thanks you, even if they don't ever know you fixed this…

rjbs (approved):

PR #170, opened 2022-01-07
replace timers to clear caches with just-in-time reaping

rjbs:
My take: no reason to leave things sitting on the event loop all the time.  Instead, we'll let things wait to get cleaned until the next read.  In theory, some more memory might get left committed, but it's tiny hash entries in all cases.


mmcclimon (approved):
This seems fine. I'll need to rewrite #145, but it'll probably be easier now! I'm not sure why I used timers in the first place, but surely it was just "we have a loop and a means of doing things periodically, might as well just use it."

PR #172, opened 2022-01-10
Linear: no creation of urgent issue w/o assignees

RashaMou:
Urgent issues must have an assignee, or we won't let you create them.

### DIFF
--- a/lib/Synergy/Reactor/Linear.pm
+++ b/lib/Synergy/Reactor/Linear.pm
@@ -429,6 +429,11 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
         # change.
         my $id  = $query_result->{data}{issueCreate}{issue}{identifier};
         my $url = $query_result->{data}{issueCreate}{issue}{url};
+        my $priority = $query_result->{data}{issueCreate}{issue}{priority};
+        my $assignee = $query_result->{data}{issueCreate}{issue}{assignee};
+        if ($priority == 1 && !defined $assignee) {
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/Linear.pm
```diff
@@ -429,6 +429,11 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
         # change.
         my $id  = $query_result->{data}{issueCreate}{issue}{identifier};
         my $url = $query_result->{data}{issueCreate}{issue}{url};
+        my $priority = $query_result->{data}{issueCreate}{issue}{priority};
+        my $assignee = $query_result->{data}{issueCreate}{issue}{assignee};
+        if ($priority == 1 && !defined $assignee) {
+          return $event->error_reply("You can't create an urgent issue with no assignee!");
+        }
```
rjbs, line 436: The code now says this:
1. make a plan
2. execute the plan, creating the issue
3. afterward, if the issue has no assignee but is urgent, declare that step 2 is forbidden

So, this won't do.  I apologize it took me so long to get here.

You should probably be examining the plan rather than the result of executing it.

rjbs (changes requested):
Changes described in comment.

lib/Synergy/Reactor/Linear.pm
RashaMou: makes sense! closing this and doing this work from within the linear client.

PR #174, opened 2022-01-14
LinearNotification: updates to slack notification message

RashaMou:
A couple of small fixes:

    - The notification message specifies if the escalated issue was
      created from Zendesk or updated to escalation in Linear

    - The notification message was incorrectly identifying the person who
      moved a issue to escalation as the creator of the issue, rather than
      the person who added the support blocker label. Linear does not give
      us an easy way to determine who the actor was, so we're removing this
      bit of info for now.

### DIFF
--- a/lib/Synergy/Reactor/LinearNotification.pm
+++ b/lib/Synergy/Reactor/LinearNotification.pm
@@ -141,11 +141,14 @@ sub http_app ($self, $env) {
           return Future->done($by_id{$payload->{data}->{creatorId}}->{displayName} // 'unknown');
         })->then(sub ($who) {
           my $desc = $was_create ? 'New task created for' : 'Existing task moved to';
+          my $app = $was_create ? 'Zendesk' : 'Linear';
+          $who = 'someone' unless $was_create;
 
           my $text = sprintf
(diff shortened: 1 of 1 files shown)


RashaMou:
@wolfsage I removed the whole api call since we weren't using it. Is this ok (and did I do it correctly!) or would you prefer I just commented out the $who for now?

wolfsage:
This seems mostly fine. We lose who created it when moved from Zendesk, but if you don't care, that's fine.

Someone *could* create a blocker in linear, and we'll say zendesk, but so what?

My only complaint is that you moved surrounding code but did not fix up the white spacing - could you do so please?

Thanks!

RashaMou:
It completely slipped my mind that we would probably still want the creator when the task is first made in Zendesk! I switched it back to how it was and changed the $who to 'someone' if it's an update. 

>> Someone could create a blocker in linear, and we'll say zendesk, but so what?
Yeah, totally no biggie, I just changed it because I was already there.

wolfsage:
nice, wfm! I would adjust the commit message to say why we are amending the message though

wolfsage (approved):

PR #175, opened 2022-01-14
Update Zendesk when Linear issue changes

RashaMou:
When a support blocker linear issue is updated with a new comment, an
internal comment is posted on its associated zendesk ticket and the
ticket status is changed to "open".


RashaMou:
@alh I took a first stab at the linear zendesk integration. I stole most of the code from LinearNotification.pm to get the http_app up and running and then just replaced the webhook response with my own stuff.

I am not completely sure I understand how http_app is set up/ how it works/ how to test the webhook (I know how it works in theory), so if you have some time to zoom or something at some point I would appreciate it!

wolfsage:
This mostly seems right, but probably what we want to do instead is having the one synergy webhook listener (LinearNotification.pm) and then multiple/different actions can happen in there when processing the webhook

As to testing, yeah we can talk in slack / zoom, up to you!

RashaMou:
@wolfsage moved the functionality into LinearNotification. I will hit you up on slack for a convo on testing!

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -171,6 +219,14 @@ sub http_app ($self, $env) {
     }
   }
 
+  # Action Zendesk ticket if comments are made on linear issues
+  my $made_comment = $payload->{type} eq 'Issue comments'
```
wolfsage, line 223: I'm not sure where you got this payload type from. Isn't it `Comment`, not `Issue comments`?

lib/Synergy/Reactor/LinearNotification.pm
```diff
@@ -80,6 +80,53 @@ has name => (
...
+
+    $self->zendesk->ticket_api->add_comment_to_ticket_f($1, {
+        body => "Issue updated with a comment: $payload->{url}",
+        public => \0,
+      })->else(sub ($err, @) {
+        $Logger->log([ "something went wrong posting to Zendesk: %s", $err ]);
+        return Future->done;
+      })->retain;
+
+    $self->zendesk->ticket_api->update_by_zendesk_id_f($1, {
```
wolfsage, line 121: I wonder if these two updates can possibly compete/block each other? It might be worth it to chain the status change to open once the add_comment_to_ticket finishes

wolfsage (commented):
I'm not sure how this ever worked. I don't think "issue comments" is a valid type. Have you tested this all?

lib/Synergy/Reactor/LinearNotification.pm
RashaMou: yeah, it's `Comment`, my mistake!

PR #179, opened 2022-02-11
pager duty: add an alias 'alerts' map to 'incidents'

Davis-A:
To often i find myself forgetting that the command is incidents and type
alerts.  Now i won't have that problem


mmcclimon:
You had `/^incidents|alerts\s*$/`, which wasn't quite right; because you needed parens for this to work, or the `^` and `$` don't do what you think.

That is: `/^foo|bar\s$/` matches "foo" and "bar", but also "football" (starts with foo) and "a guy walks into a bar" (ends with bar). What you needed here instead was `/^(incidents|alerts)\s*$/`. Seems hardly worth the round trip, so I just amended and merged this. Thanks!

PR #181, opened 2022-04-07
Linear: add slack link to linear issue

RashaMou:
When creating a linear issue from slack, add the slack link to the issue
so we can see any background discussion around the task creation


RashaMou:
@mmcclimon I'm not sure how to get this onto slack to actually test it

PR #190, opened 2022-10-06
Linear Reactor: comment on Linear issues

RashaMou:
Synergy will react to "comment on plu-xxx: comment" by adding a comment to the given issue

### DIFF
--- a/lib/Synergy/Reactor/Linear.pm
+++ b/lib/Synergy/Reactor/Linear.pm
@@ -578,7 +578,7 @@ command comment => {
           my $slack = sprintf("I added that comment to <%s|%s>.", $url, $issue_ident);
           return $event->reply($text, { slack => $slack });
         }
-        $Logger->log("Err: $res");
+        $Logger->log("error trying to create a comment on $issue_ident: $res");
         return $event->error_reply("Sorry, something went wrong and I can't say what!");
       })
     })
--- a/lib/Synergy/Reactor/Linear.pm
+++ b/lib/Synergy/Reactor/Linear.pm
@@ -552,6 +552,39 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
   $self->_with_linear_client($event, $code);
 }
 
+command comment => {
+  help => "*comment on `ISSUE`: `comment`*: add a comment to a linear issue",
+} => sub ($self, $event, $rest) {
+  $event->error_reply("I don't know what you want me to comment on.") unless $rest;
+  my ($issue_ident, $comment) = $rest =~ /
+    ^
+    (?:on\s+)?
+    ([a-z]+-[0-9]*)
+    :?
+    \s+
+    (.*)
+    \z
+  /ix;
+  $event->error_reply("I don't know what you want me to comment on.") unless $issue_ident && $comment;
+  $self->_with_linear_client($event, sub ($linear) {
+    my $issue_f = $linear->fetch_issue($issue_ident);
+    $issue_f->then(sub ($issue) {
+      my $comment_f = $linear->create_comment({issueId => $issue->{id}, body => $comment});
+      $comment_f->then(sub ($res) {
+        my $comment_id = $res->{data}{commentCreate}{comment}{id};
+        my $url = $res->{data}{commentCreate}{comment}{url};
+        if ($comment_id) {
+          my $text = sprintf("I added that comment to %s: %s.", $url, $issue_ident);
+          my $slack = sprintf("I added that comment to <%s|%s>.", $url, $issue_ident);
+          return $event->reply($text, { slack => $slack });
+        }
+        $Logger->log("Err: $res");
+        return $event->error_reply("Sorry, something went wrong and I can't say what!");
+      })
+    })
+  });
+};
+
 responder new_issue => {
   exclusive => 1,
   targeted  => 1,


RashaMou:
@rjbs synergy comment code ready for review

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,57 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+      my $comment_id = $query_result->{data}{commentCreate}{comment}{id};
+      my $url = $query_result->{data}{commentCreate}{comment}{url};
+      if ($comment_id) {
+        return $event->reply(
+          sprintf("I added that comment on %s: %s", $issue, $url),
+          {
+            slack => sprintf("I added that comment on <%s|%s>.", $url, $issue),
+          },
+        );
+      } else {
```
rjbs, line 573: "Never write `else` blocks." 🙂

The previous `if` will have returned, so no need for else.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,57 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+      }
+    });
+  };
+  if ($linear) {
+    return $code->($linear);
+  }
+  $self->_with_linear_client($event, $code);
+}
+
+responder new_comment => {
```
rjbs, line 586: This should be a command, not a responder.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,57 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
   $self->_with_linear_client($event, $code);
 }
 
+sub _handle_comment_event ($self, $event, $arg = {}) {
+  $event->mark_handled;
```
rjbs, line 556: You don't need mark_handled if the caller is a command instead of a responder.

rjbs (changes requested):

lib/Synergy/Reactor/Linear.pm
RashaMou: noted, but also this whole sub is gone now \o/

lib/Synergy/Reactor/Linear.pm
RashaMou: changed to a command - so much better!

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,32 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
   $self->_with_linear_client($event, $code);
 }
 
+command comment => {
+  help => "*comment on `ISSUE`: `comment`*: add a comment to a linear issue",
+  aliases => ["comment on"],
```
rjbs, line 557: This doesn't make sense as an alias to me.  Comment matches the first word.  The "on" is in $rest.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,32 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+} => sub ($self, $event, $rest) {
+  return unless $rest;
+  my ($issue, $comment) = $rest =~ /\s+(.*-\d*)+:?(.*)/;
```
rjbs, line 560: This should be anchored at both ends, probably, and should account for "on"`

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,32 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+  return unless $issue && $comment;
```
rjbs, line 561: At this point, in a command, we have committed to handling this command.  We must not just return, but should provide an error message instead.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,32 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+  $comment=~ s/^\s+//;
+  $self->_with_linear_client($event, sub ($linear) {
+    my $issue_id_f = $linear->fetch_issue($issue);
```
rjbs, line 564: This is fetching an issue, not an issue id. You are passing it an issue identifier.  We should put the identifier in $issue_ident, then pass that in, then store the future in $issue_f.

This is a bit nit-picky, sorry, but it's important:  Without a real type system, we have to rely on the variable names to tell us what's what.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,32 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+    $issue_id_f->then(sub ($id) {
```
rjbs, line 565: …and then $id here is $issue.  If you call it $id and then say $id->{id} some future reader's head will explode! 😄

rjbs (changes requested):
A few small things, but very close!

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,31 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
   $self->_with_linear_client($event, $code);
 }
 
+command comment => {
+  help => "*comment on `ISSUE`: `comment`*: add a comment to a linear issue",
+} => sub ($self, $event, $rest) {
+  $event->error_reply("I don't know what you want me to comment on.") unless $rest;
+  my ($issue_ident, $comment) = $rest =~ /^(?:on\s)?(.*-\d*)+:?(.*)\z/;
```
rjbs, line 559:
```suggestion
  my ($issue_ident, $comment) = $rest =~ /
    ^
    (?:on\s+)?
    ([a-z]+-[0-9]*)
    :?
    \s+
    (.*)
    \z
  /ix;
```

Lots of tiny changes made here:

1. use /x to allow newlines and space in the regex so it's more readable
2. put `+` after the space after "on" so you can say `on    CLI-123` if you hit space too long
3. replaced `.*` in the left hand side of the CLI-123 matcher with `[a-z]+` — .* is exceptionally greedy; remember that `.` would match `:` and all kinds of weird things could be parsed weird
4. replaced `\d` with `[0-9]` — this is a pet peeve of mine.  \d matches lots of things you don't want.  I have annoyingly included a small sample below
5. added `\s+` to go between issue identifier and comment so you can't say `comment cli-123it's done`

Sorry for doing "here's just code" but these are all small but worth doing, and the Socratic method was probably going to be really obnoxious here!

```
 ๖  U+0E56 THAI DIGIT SIX
 ໖  U+0ED6 LAO DIGIT SIX
 ༦  U+0F26 TIBETAN DIGIT SIX
 ၆  U+1046 MYANMAR DIGIT SIX
 ႖  U+1096 MYANMAR SHAN DIGIT SIX
 ៦  U+17E6 KHMER DIGIT SIX
 ᠖  U+1816 MONGOLIAN DIGIT SIX
 ᥌  U+194C LIMBU DIGIT SIX
 ᧖  U+19D6 NEW TAI LUE DIGIT SIX
 ᪆  U+1A86 TAI THAM HORA DIGIT SIX
```

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,31 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+  $event->error_reply("I don't know what you want me to comment on.") unless $issue_ident && $comment;
+  $comment=~ s/^\s+//;
```
rjbs, line 561: This line not needed with the regex change above.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,31 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+  $self->_with_linear_client($event, sub ($linear) {
```
rjbs, line 562: I have a branch now where I async/await all of the Linear reactor, and it's so nice.  We'll do this routine after I merge my branch.  I think you will be pleased.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,31 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+    my $issue_f = $linear->fetch_issue($issue_ident);
+    $issue_f->then(sub ($issue) {
+      my $comment_f = $linear->create_comment({issueId => $issue->{id}, body => $comment});
+      $comment_f->then(sub ($res) {
+        my $comment_id = $res->{data}{commentCreate}{comment}{id};
+        $event->reply($comment_id);
```
rjbs, line 568: This seems like debugging noise left over.  No human wants this information. 🙂

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,31 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+        my $url = $res->{data}{commentCreate}{comment}{url};
+        if ($comment_id) {
+          my $text = sprintf("I added that comment to <%s|%s>.", $url, $issue_ident);
+          return $event->reply($text, { slack => $text });
```
rjbs, line 572: Here, you're sending Slack-formatted text (the `<link|text>`) to plaintext channels like SMS.  You should have two different strings to send: one plain, one Slack.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,31 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+        }
+        return $event->error_reply("Sorry, something went wrong and I can't say what!");
```
rjbs, line 574: In this case, you should $Logger log the $res, so the logs show what happened.

rjbs (changes requested):
Very close, mostly a change for you to review and apply, 1-2 small nits.  Deploy Monday?

Also, I think you need to add some vertical whitespace between paragraphs and maybe wrap some long lines.

I am going to use the hell out of this feature.

lib/Synergy/Reactor/Linear.pm
RashaMou: ooo yes! excited for that.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -552,6 +552,39 @@ sub _handle_creation_event ($self, $event, $arg = {}) {
...
+      my $comment_f = $linear->create_comment({issueId => $issue->{id}, body => $comment});
+      $comment_f->then(sub ($res) {
+        my $comment_id = $res->{data}{commentCreate}{comment}{id};
+        my $url = $res->{data}{commentCreate}{comment}{url};
+        if ($comment_id) {
+          my $text = sprintf("I added that comment to %s: %s.", $url, $issue_ident);
+          my $slack = sprintf("I added that comment to <%s|%s>.", $url, $issue_ident);
+          return $event->reply($text, { slack => $slack });
+        }
+        $Logger->log("Err: $res");
```
rjbs, line 581:
```suggestion
        $Logger->log("error trying to create a comment on $issue_ident: $res");
```

rjbs (approved):

PR #192, opened 2022-10-08
Linear: convert to Async/Await

rjbs:
This converts the Linear reactor to async/await.  The whole thing can be further updated and improved, but this already seems like a good update.  I have tested much of this by hand already.

### DIFF
--- a/lib/Synergy/Reactor/Linear.pm
+++ b/lib/Synergy/Reactor/Linear.pm
@@ -585,8 +585,11 @@ async sub _handle_creation_event ($self, $event, $arg = {}) {
   my $ersatz_text = $arg->{ersatz_text};
 
   my $code = async sub ($linear) {
-    my $slack_link = $event->event_uri;
-    my $text = $event->text . "\n\ncreated at: $slack_link";
+    my $text = $event->text;
+
+    if (my $slack_link = $event->event_uri) {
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/Linear.pm
```diff
@@ -11,9 +11,10 @@ with 'Synergy::Role::Reactor',
        expandos => [ 'issue' ],
      };
 
-use experimental qw(signatures lexical_subs);
+use experimental qw(signatures lexical_subs try);
 use namespace::clean;
 
+use Future::AsyncAwait;
```
mmcclimon, line 17: We should make sure this ends up in the cpanfile/makefile/wevs.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -11,9 +11,10 @@ with 'Synergy::Role::Reactor',
        expandos => [ 'issue' ],
      };
 
-use experimental qw(signatures lexical_subs);
+use experimental qw(signatures lexical_subs try);
```
mmcclimon, line 14: oh hey I didn't realize synergy's perl had been upgraded.

mmcclimon (approved):
I think this is fine; if it doesnt work, you can fix it.

PR #194, opened 2023-03-02
inabox: Make box size as a config option

Davis-A:
Dumb little change.  I don't touch Synergy very often these days and was fiddling with it anyway so I thought why not 🤷


wolfsage:
Sure!

wolfsage (approved):

PR #195, opened 2023-03-02
Add emergency reactor

RashaMou:
The  commands returns the emergency procedures outlined in https://www.notion.so/fastmailteam/I-have-an-urgent-problem-040fb410668f4e92beb43aa40695ce65


RashaMou:
@rjbs per yassar's request

PR #197, opened 2023-06-02
Add a very basic 'search ...' to Linear Reactor

wolfsage:
This searches tickets and comments in the same way the UI does when using the Advanced Search interface


wolfsage:
This code was very much "get it working while under duress" so there's definitely some stuff that could be refactored / written more cleanly. If any of it bothers you enough let me know :)

wolfsage:
The bolding stuff doesn't work if we're trying to bold a username because it becomes `@*alh*` which slack does nothing with. I can maybe work around that...

wolfsage:
Maybe could use https://api.slack.com/messaging/composing/layouts#stack_of_blocks text blocks with 'verbatim' and 'mrkdwn' types to mix/compose the output. Ugh

rjbs:
Pretty sure the blocks will become, like, *blocks*.  So, breaks between them.  No good for inline formatting.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -21,6 +21,8 @@ use Synergy::CommandPost;
 use Synergy::Logger '$Logger';
 use Synergy::Util qw(bool_from_text reformat_help);
 
+use POSIX qw(ceil);
```
rjbs, line 24: I hate that this is in POSIX.  But that isn't your fault.

As far as I know.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+
+    Searches linear the same way advanced search does (but only for
+    tickets/comments).
+
+    Arguments:
+
+      * TEXT: The string(s) to search
+      * /page N: Which page of results to retrieve, if there's more than 1
+      * /count N: Return N results. Default is 10. Any more will be a private response
+      * /closed: Include tasks in the canceled/completed states
```
rjbs, line 335:
```suggestion
    • TEXT: The string(s) to search
    • /page N: Which page of results to retrieve, if there's more than 1
    • /count N: Return N results. Default is 10. Any more will be a private response
    • /closed: Include tasks in the canceled/completed states
```

This suggestion matches formatting with other help lists in Linear reactor, like "new_issue"

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+    EOH
+} => async sub ($self, $event, $rest) {
+  my $wpage = 1;
+  if ($rest =~ s{\s/page\s+(\d+)}{}i) {
```
rjbs, line 339: I know, I know, but:  you should put `/a` on these, because you don't want somebody to say `/pages ६` and cause some unhandled error.

rjbs: Also do you want `\s*` after the captured digit string?  I think you might.  But maybe it doesn't matter in the final analysis.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+  if ($rest =~ s{\s/closed\b}{}i) {
+    $filter_state_how = 'in';
+  }
+
+  $event->mark_handled;
+
+  my $linear = $self->_linear_client_for_user($event->from_user);
+  unless ($linear) {
+    return await $event->reply("You don't have a linear token...");
+  }
```
rjbs, line 359: Often for this kind of code, we use `_with_linear_client` — maybe you want to use that here.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+
+  my $gen = sub ($pager, @rest) {
+    unless (
+      @rest == 0
+      ||
+      @rest == 2 && ($rest[0] eq 'after' || $rest[0] eq 'before')
+    ) {
+      die "confused! [@rest]";
+    }
```
rjbs, line 368: Ugh, I'm sorry that you had to write this.  It's my fault.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+              state => [ qw(name type) ],
+            ],
+          },
+        ],
+      },
+    );
+
+    # meh. selection_set wants k => v pairs, but this totalCount thing is on
+    # its own. -- alh, 2023-06-02
+    my $query = "query {\n" . $selection->as_string("  ") . "\n}\n";
```
rjbs, line 404: I believe you want:
```
select => [
  totalCount =>
  pageInfo => { ... },
  nodes => { ... }
]
```

That is:  I believe it takes an optlist (a la Data::OptList).

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+
+    $query =~ s/pageInfo/totalCount\n    pageInfo/;
+
+    return $query;
+  };
+
+  my $xtract  = sub { $_[0]{data}{searchIssues} };
```
rjbs, line 411: I'd use a signature.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+  my $cpage = 1;
+
+  while ($wpage > $cpage) {
+    unless ($query_result->has_next_page) {
+      return await $event->reply("You requested page $wpage, out of range (max: $cpage)");
+    }
+
+    $cpage++;
+
+    $query_result = await $query_result->next_page;
```
rjbs, line 431: Probably the polite way to do this is something else, like embed a link in the message that triggers Synergy to fetch the second page by re-providing the stored cursor…

…but this should be fine unless we get rate limited from people saying page:4

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+  }
+
+  my $text = my $slack = "*Your results for $rest...*\n";
```
rjbs, line 434:
```suggestion
  my $text = my $slack = "*Your results for «$rest»:*\n";
```

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+    my $state = $node->{state}{name};
+    my $title = $node->{title};
+
+    my $icon = $self->_icon_for_issue($node);
+
+    my $snippet;
+    my $highlights;
+    my $is_title;
+
+    PATH: for my $path ([qw(title)], [qw(comment body)], [qw(description)]) {
```
rjbs, line 449: Cute. :)

lib/Synergy/Reactor/Linear.pm
```diff
@@ -318,6 +320,208 @@ async sub _handle_search ($self, $event, $arg) {
...
+
+      $snippet = $t->{snippet};
+      $highlights = $t->{highlights};
+
+      last;
+    }
+
+    unless ($highlights) {
+      warn "Oops, unhandled snippet response: " . Dumper($node);
+      warn Dumper $node;
```
rjbs, line 474: This should probably use $Logger.

rjbs (changes requested):
Looks pretty good!

lib/Synergy/Reactor/Linear.pm
wolfsage: The rate limiting is pretty generous and done by points usage. Each page for a simple query is about 8 points, and you get 250,000 an hour, I think we're good

lib/Synergy/Reactor/Linear.pm
wolfsage: :)

lib/Synergy/Reactor/Linear.pm
wolfsage: Fixed

lib/Synergy/Reactor/Linear.pm
wolfsage: Fixed (Half the code in hear does what I was doing which is where I got it from :))

lib/Synergy/Reactor/Linear.pm
wolfsage: I didn't! I copied this mostly from elsewhere. And now I've undone it all because it turns out I can just us do_paginated_query!

lib/Synergy/Reactor/Linear.pm
wolfsage: Fixed (requires https://github.com/fastmail/Linear-Client/pull/37)

lib/Synergy/Reactor/Linear.pm
wolfsage: Fixed

wolfsage (commented):
Updated a bunch and cleaned up. This needs https://github.com/fastmail/Linear-Client/pull/37 and also Feature::Compat::Defer (just for debugging... so we could remove that bit if we want)

wolfsage:
Will squash all commits before merging since history is a mess

rjbs:
@wolfsage I pushed a commit to update prereqs.  I merged and installed the new client on synergy.  I installed defer there too.

Branch looks good, ship it.

PR #198, opened 2023-06-28
Synergy::Hub: Track in-flight http requests

wolfsage:
This makes it so that Synergy::Channel::Test's is_exhausted can take into account whether or not a reactor is waiting on an external HTTP service before responding.

Without this, `$channel->is_exhausted` would return true, even if a reactor hadn't had a chance to respond yet.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -2,7 +2,7 @@
 use strict;
 use warnings;
 
-use 5.034000;
+use 5.032000;
 
 use ExtUtils::MakeMaker 6.78;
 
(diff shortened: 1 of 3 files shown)


wolfsage:
This also make t/inabox.t way more reliable.

The test will still heisenfail though because it chains HTTP requests one after another completes, which these changes cannot account for (as in flight requests will go to 0 then back up again) and the is_exhausted check can hit in that window.

Really I think like with "mark_handled" we want "mark_completed" and then is_exhausted can instead check for any reactors that claim to have  in flight work going...

rjbs:
Seems reasonable to me, nice!

rjbs (approved):

PR #200, opened 2023-07-10
restore compatibility with legacy version v5.32

rjbs:


lib/Synergy/Reactor/Linear.pm
```diff
@@ -748,12 +747,19 @@ command agenda => {
       } else {
         my ($assignee_id, $team_id);
 
-        try {
+        my $error_reply;
```
wolfsage, line 750: nit: not used

lib/Synergy/Reactor/Linear.pm
```diff
@@ -871,12 +877,16 @@ async sub _handle_creation_event ($self, $event, $arg = {}) {
     $text =~ s/\A> >/>>/;
 
     # XXX: I do not like our current error-returning scheme. -- rjbs, 2021-12-10
-    my $plan;
-    try {
+    my ($plan, $error_reply);
```
wolfsage, line 880: nit: not used (error_reply)

lib/Synergy/Reactor/RememberTheMilk.pm
```diff
@@ -217,11 +226,16 @@ command milkauth => {
       …and then tell me `milkauth complete`
       END
 
-      return await $event->private_reply($text, { slack => $text });
-    } catch ($fail) {
-      $Logger->log([ "failed to make start auth: %s", $fail ]);
-      return await $event->reply("Sorry, something went wrong!");
+      return $event->private_reply($text, { slack => $text });
```
wolfsage, line 229: not a return await which seems weird

lib/Synergy/Reactor/Zendesk.pm
```diff
@@ -249,18 +254,23 @@ async sub ticket_report ($self, $who, $arg = {}) {
   }
 
   my $res;
-  try {
+  my $to_return;
```
wolfsage, line 257: nit: unused

lib/Synergy/Reactor/RememberTheMilk.pm
rjbs: This returns it from the eval, after which we `return await` it.  We need to return out of the eval.

wolfsage: Ah! Okay

PR #201, opened 2023-07-11
Zendesk reactor: skip getting zendesk tickets if user has no zendesk account

RashaMou:
@wolfsage I am not sure how to test this locally.

### DIFF
--- a/lib/Synergy/Reactor/Zendesk.pm
+++ b/lib/Synergy/Reactor/Zendesk.pm
@@ -21,6 +21,7 @@ use Synergy::Logger '$Logger';
 use Time::Duration qw(ago);
 use Try::Tiny;
 use Zendesk::Client;
+use Synergy::Logger '$Logger';
 
 __PACKAGE__->add_preference(
   name      => 'staff-email-address',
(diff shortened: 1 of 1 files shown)


RashaMou:
@wolfsage this is ready for review

wolfsage:
This should log if we don't find a user, and also log if there's some kind of exception. This way we can see why reports might not fire for a user and not have to guess

wolfsage (changes requested):

lib/Synergy/Reactor/Zendesk.pm
```diff
@@ -253,6 +254,26 @@ async sub ticket_report ($self, $who, $arg = {}) {
...
+  if ($@) {
+    my $error = $@;
+    $Logger->log([
+      "Error while attempting to get Zendesk user for %s: %s",
+      $email,
+      $error
+    ]);
+  };
+
+  unless ($user_ok) {
```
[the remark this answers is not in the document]
RashaMou: yes, it does, thanks

lib/Synergy/Reactor/Zendesk.pm
```diff
@@ -253,11 +254,31 @@ async sub ticket_report ($self, $who, $arg = {}) {
...
+                ]);
+    return;
+  };
+
   my $res;
   my $ok = eval {
     $res = await $self->zendesk_client
                       ->user_api
-                      ->get_by_email_no_fetch($email)
+                      ->$user
```
wolfsage, line 281: This change makes no sense. Really what you'd do at this point is `$res = $user->scoped_client->make_request_f...` probably. Maybe we pair on this this week and do it live on synergy and see what happens :)

RashaMou: Yeah that would be great, there's something I'm not quite getting here!

PR #202, opened 2023-08-21
Allow more formal bot address, by comma

tjwds:


rjbs (changes requested):
This isn't wrong, but better would probably be to make them use the same expression.  Why is one spaced differently?  Why does only one quotemeta?  If they're the same, easier for someone to extract them to common code later (or sooner)!

tjwds:
@rjbs Where would you put that expression?  Util?

rjbs:
@tjwds Yeah, Util is the real fruit salad of functions.  Alternatively, it could go into Synergy::Role::Channel, which might not be a bad idea.

rjbs:
@tjwds Last(?) nit:  Commit style is "CLASS: message" so maybe:

```
Channel role: add a common way to get the thing
Discord and Slack channels: use the common thing foo_bar
```

rjbs:
All tests pass.  It would definitely be easy to write tests for this using Synergy::Channel::Test

rjbs (approved):
👍🏽

rjbs:
This broke things.  I don't know why yet. Let's fix it tomorrow. :)

PR #203, opened 2023-09-15
convert In-a-Box reactor to new DO API client

rjbs:
This branch extracts much of the DigitalOcean API code from the InABox reactor and into a simple reusable library.  Probably that library should be split out and made more comprehensive and a bit more generic.  That can happen later.

The real drive here was to make that library usable outside Synergy, but to make that easy to understand, some other changes were useful to make along the way.  They are…

* the InABox reactor now uses async/await, simplifying the code
* the InABox reactor has been converted to CommandPost, the Reactor role of the future
* I've introduced Synergy::X, a generic Throwable for Synergy

Synergy::X is intended to be used to help signal failure in Synergy, most especially in async/await methods, and most most especially when you want to throw an error down in some async/await chain that can bubble up as a user-visible error report later.


lib/Dobby/Client.pm
```diff
@@ -0,0 +1,273 @@
...
+  $path .= "&tag_name=$arg->{tag}" if $arg->{tag};
+
+  my $droplets_data = await $self->json_get($path);
+
+  # TODO Obviously, this should lazily fetch etc.
+  if ($droplets_data->{links}{pages}{forward_links}) {
+    Carp::cluck("Single-page fetch did not find all droplets!");
+  }
+
+  return $droplets_data->{droplets}->@*;
```
wolfsage, line 211: This'll crash if we didn't get any droplets. Should we validate the json first for future debugging of errors?

wolfsage: I guess this (and a bunch of my earlier points) come down to: "Do we trust the DO API to remain stable and return was it says it'll return on success.". (My gut says no.)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -219,177 +172,151 @@ sub _determine_version_and_tag ($self, $event, $switches) {
...
+    $event->error_reply(
+      "Box was created, but now I can't find it! Check the DigitalOcean console and maybe try again."
+    );
+  }
+
+  # Add it to the relevant project. If this fails, then...oh well.
+  # -- ?
+  #
+  # I'm putting the result in a variable, because not doing so led to some
+  # weird-o error from AsyncAwait.  I'll ask LeoNerd about it. -- rjbs,
```
wolfsage, line 256: amazing

lib/Synergy/Reactor/InABox.pm
```diff
@@ -399,79 +326,58 @@ sub _handle_power ($self, $event, $action, $tag = undef) {
...
 
-  $self->_get_droplet_for($event->from_user, $tag)
-    ->then(sub ($maybe_droplet) {
-      unless ($maybe_droplet) {
-        return Future->fail(
-          "That box doesn't exist: " . $self->_box_name_for($event->from_user, $tag),
-          'stop-processing',
-        );
-      }
+  my $droplet = await $self->_get_droplet_for($event->from_user, $tag);
```
wolfsage, line 336: We should check that the box exists here like the code above used to, I think

lib/Synergy/Reactor/InABox.pm
```diff
@@ -573,43 +455,34 @@ sub _format_droplet ($self, $droplet) {
...
 
-sub _get_snapshot ($self, $version) {
-  return $self->_do_request(GET => '/snapshots?per_page=200')
-    ->then(sub ($data) {
-      my ($snapshot) = sort { $b->{name} cmp $a->{name} }
-                       grep { $_->{name} =~ m/^fminabox-\Q$version\E/ }
-                       $data->{snapshots}->@*;
+async sub _get_snapshot ($self, $version) {
+  my $dobby = $self->dobby;
+  my $data  = await $dobby->json_get('/snapshots?per_page=200');
```
wolfsage, line 460: Looks like you added `sub _get_snapshot` to dobby that is this same method, but nothing's using it. Would it be better to make that `sub get_snapshot` and use `$self->dobby->get_snapshot...` instead?

lib/Synergy/X.pm
```diff
@@ -0,0 +1,54 @@
+use v5.32.0;
+use warnings;
+package Synergy::X;
+
+use Moose;
+with 'Throwable',
+     'Role::Identifiable::HasIdent';
```
wolfsage, line 7: Does cpanfile need an update for this or is something else already pulling it in?

wolfsage (commented):
I have complained a lot.

Overall I think Dobby an the async stuff make this all much easier to read and reason about. Good stuff!

lib/Synergy/Reactor/InABox.pm
rjbs: Actually, I believe I solved this and so will remove that variable and the comment and double-check.

lib/Synergy/Reactor/InABox.pm
rjbs: Yes, good catch, thanks.

PR #204, opened 2023-09-22
convert preference properties to async/await (and associated work)

rjbs:
The big goal here was "allow a preference describer to be async, so it can look up an opaque identifier in a remote API without locking up the whole bot".  To accomplish that, or as closely related work…

* the HasPreferences system uses async/await and made validator, describer, and after_set async
* the describer callback gets the reactor as first arg
* all reactors using those callbacks were updated to use the new semantics
* the Preferences reactor was updated to match, and to use CommandPost and async/await
* the PagerDuty reactor was updated to use CommandPost and async/await (see below)
* the Linear reactor was updated to use the Linear API to describe the default team setting

The PagerDuty reactor was pretty extensively updated.  It didn't need to be, strictly, but while doing this work, I found a few parts sort of tedious to update because of its extensive use of sequenced futures.  Since I knew there was at least one recent request for changes to this reactor, I thought that the change to async/await would be a good bit of work to throw on the pile, to make that request easier for someone to grant.

### DIFF
--- a/lib/Synergy/Hub.pm
+++ b/lib/Synergy/Hub.pm
@@ -196,8 +196,7 @@ sub handle_event ($self, $event) {
         ]);
 
         if ($args[0] isa 'Synergy::X' && $args[0]->is_public) {
-          $event->reply($args[0]->message);
-          return
+          return $event->reply($args[0]->message);
         }
 
         $event->error_reply("My $rname reactor crashed (in the background) while handling your message.  Sorry!");
--- a/lib/Synergy/Reactor/Clox.pm
+++ b/lib/Synergy/Reactor/Clox.pm
@@ -12,6 +12,7 @@ use Synergy::CommandPost;
 use experimental qw(signatures lexical_subs);
 
 use DateTime;
+use Future::AsyncAwait;
 use List::Util qw(first uniq);
 use Synergy::Util qw(parse_date_for_user);
 use Time::Duration::Parse;
@@ -25,8 +26,8 @@ __PACKAGE__->add_preference(
   name        => 'include-aelt',
   help        => "Whether or not to include AELT in the output",
   description => "Whether or not to include AELT in the output",
-  describer   => sub ($value) { $value ? 1 : 0 },
-  validator   => sub ($self, $value, $event) {
+  describer   => async sub ($value) { $value ? 1 : 0 },
+  validator   => async sub ($self, $value, $event) {
     return(($value ? 1 : 0), undef); # Should write a generic bool/yes/no.
   },
   default     => 0,
--- a/lib/Synergy/Reactor/Clox.pm
+++ b/lib/Synergy/Reactor/Clox.pm
@@ -26,7 +26,7 @@ __PACKAGE__->add_preference(
   name        => 'include-aelt',
   help        => "Whether or not to include AELT in the output",
   description => "Whether or not to include AELT in the output",
-  describer   => async sub ($value) { $value ? 1 : 0 },
+  describer   => async sub ($self, $value) { $value ? 1 : 0 },
   validator   => async sub ($self, $value, $event) {
     return(($value ? 1 : 0), undef); # Should write a generic bool/yes/no.
   },
--- a/lib/Synergy/Reactor/DiagnosticEvents.pm
+++ b/lib/Synergy/Reactor/DiagnosticEvents.pm
@@ -21,7 +21,7 @@ command 'diag-die' => {
 command 'diag-die-x' => {
   help => '*diag-die-x*: reacts by throwing a public Synergy::X exception',
 } => async sub ($self, $event, $rest) {
-  Synergy::X->throw_public("This exception was caused on purpose via Synergy::X.");
+  Synergy::X->throw_public("This exception was caused on purpose via a Synergy::X thrown by diag-die-x.");
 };
 
 1;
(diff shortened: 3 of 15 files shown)


lib/Synergy/Hub.pm
```diff
@@ -195,11 +195,21 @@ sub handle_event ($self, $event) {
           "@args", # stupid, but avoids json serialization guff
         ]);
 
+        if ($args[0] isa 'Synergy::X' && $args[0]->is_public) {
+          $event->reply($args[0]->message);
+          return
```
wolfsage, line 200: nit: bare return

lib/Synergy/Reactor/DiagnosticEvents.pm
```diff
@@ -0,0 +1,27 @@
...
+command 'diag-die' => {
+  help => '*diag-die*: reacts by throwing a string exception',
+} => async sub ($self, $event, $rest) {
+  die "This exception was caused on purpose by diag-die.";
+};
+
+command 'diag-die-x' => {
+  help => '*diag-die-x*: reacts by throwing a public Synergy::X exception',
+} => async sub ($self, $event, $rest) {
+  Synergy::X->throw_public("This exception was caused on purpose via Synergy::X.");
```
wolfsage, line 24: nit: via diag-die-x but this is fine

lib/Synergy/Reactor/Clox.pm
```diff
@@ -25,8 +26,8 @@ __PACKAGE__->add_preference(
   name        => 'include-aelt',
   help        => "Whether or not to include AELT in the output",
   description => "Whether or not to include AELT in the output",
-  describer   => sub ($value) { $value ? 1 : 0 },
-  validator   => sub ($self, $value, $event) {
+  describer   => async sub ($self, $value) { $value ? 1 : 0 },
+  validator   => async sub ($self, $value, $event) {
```
wolfsage, line 30: So this is a change in the expected API. validator/describer subs must be async now I assume or things fail. That's fine, I think we basically treat this project as having no downstream users, just wanted to make sure we considered this!

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -169,146 +171,377 @@ sub start ($self) {
...
-      },
-      allow_empty_help => 1,  # provided by maint-query
-    },
+# "start /force" is not documented here because it will mention itself to the
+# user when needed -- rjbs, 2020-06-15
+help maint => reformat_help(<<~"EOH");
+  Conveniences for managing PagerDuty's "maintenance mode", aka "silence all
+  the alerts because everything is on fire."
+
+  • *maint status*: show current maintenance state • *maint start*: enter
```
wolfsage, line 180: Huh, will these bullets get reformatted with newlines properly?

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -169,146 +171,377 @@ sub start ($self) {
...
+    return [ ] if $text =~ m{\Ademaint\s*\z}ni;
+    return;
+  },
+} => async sub ($self, $event) {
+  $event->mark_handled;
+
+  unless ($self->is_known_user($event)) {
+    return await $event->error_reply("I don't know you, so I'm ignoring that.");
+  }
+
```
wolfsage, line 283: nit: Mildly surprised we don't require /force for non-oncall to demaint

lib/Synergy/Reactor/Preferences.pm
```diff
@@ -7,154 +7,117 @@ use Try::Tiny;
...
+  • *clear my `PREF`*: unset a preference and go back to the default
 
-sub handle_set ($self, $event) {
+  If you're an administrator, you can also…
+
+  • *set `USER`'s `PREF` to `VALUE`*: set another user's preferences
+  • *clear `USER`'s `PREF`*: unset some other user's preference
+  EOH
+
+command set => { # allow_empty_help => 1,  # provided by preferences above
```
wolfsage, line 41: What's up with the double comments?

wolfsage (approved):

lib/Synergy/Hub.pm
rjbs: Updated to return the reply future.

lib/Synergy/Reactor/Clox.pm
rjbs: Exactly so.

lib/Synergy/Reactor/PagerDuty.pm
rjbs: No this was an error, and has been fixed!

lib/Synergy/Reactor/PagerDuty.pm
rjbs: I believe I'm just carrying forward existing behavior!

lib/Synergy/Reactor/Preferences.pm
rjbs: This is maybe my next thing to fix.  Last year, Michael added something so that reactors with commands that had no help would log this at start up, with `allow_empty_help` as an escape clause.  Great feature!

Unfortunately, it didn't make it into CommandPost.  I want to re-add it, and the above is a bit of a note-to-self to mark these allowed empty.

PR #206, opened 2023-09-25
add documentation to the CommandPost system

rjbs:


lib/Synergy/CommandPost.pm
```diff
@@ -2,9 +2,160 @@ use v5.32.0;
...
+boundary.  Commands are exclusive reactions, so they won't work if any other
+exclusive reaction matches.
+
+The subroutine provided for the command is invoked with the Synergy::Event that
+triggered it, plus the arguments to the command (for which, see below).
+
+The following options can be passed to the command declarator:
+
+=for :list
+* help - the help text for the command
```
wolfsage, line 54: nit: no newline after a `=` block always feels weiiiiiiiiiird

lib/Synergy/CommandPost.pm
```diff
@@ -2,9 +2,160 @@ use v5.32.0;
...
+* aliases - an arrayref of other names for the command; only the real name
+  appears in the help index
+* parser - a coderef to parse the rest of the user input
+
+If no C<parser> is provided, then the user's text after the initial command is
+passed as the third argument to the method.  In other words, with no parser,
+"eject all pods" calls C<< $command_sub->($reactor, $event, "all pods") >>.
+
+If a parser I<is> provided, it will be called with the reactor as the first
+argument, the default value (like "all pods") as the second, and the event as
```
wolfsage, line 64: I'm confused by 'default value' here. What does that mean?

wolfsage (approved):

lib/Synergy/CommandPost.pm
rjbs: `=for` is special and requires no newline be present.

lib/Synergy/CommandPost.pm
rjbs: It means "the text after the command" and I've updated it to say so.

PR #211, opened 2023-11-03
Fix syntax error

lerlacher-fm:
I stepped into several traps and broke the build. This fixes it.


Davis-A (approved):

rjbs:
There exists a test suite.  Please run it before pushing.  It includes a "does every file compile" test.

PR #212, opened 2023-11-03
Inabox setup snippet

Davis-A:
I really dislike getting a raw text file reply when something goes wrong!

### DIFF
--- a/lib/Synergy/Reactor/GitLab.pm
+++ b/lib/Synergy/Reactor/GitLab.pm
@@ -1109,6 +1109,27 @@ sub mr_report ($self, $who, $arg = {}) {
   });
 }
 
+async sub post_gitlab_snippet ($self, $payload) {
+  my $res = await $self->hub->http_client->do_request(
+    method => 'POST',
+    uri    => $self->api_uri . '/v4/snippets',
+    headers => {
+      'PRIVATE-TOKEN' => $self->api_token,
+    },
+
+    content_type => 'application/json',
+    content      => encode_json($payload),
+  );
+
+  unless ($res->is_success) {
+    die "error creating snippet on GitLab: " . $res->as_string;
+  }
+
+  my $json = $res->decoded_content(charset => undef);
+  my $snippet_metadata = decode_json($json);
+  return $snippet_metadata->{web_url};
+}
+
 __PACKAGE__->add_preference(
   name      => 'user-id',
   validator => async sub ($self, $value, @) {
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -422,9 +422,7 @@ async sub _setup_droplet ($self, $event, $droplet, $key_file, $args = []) {
     return await $event->reply("In-a-Box ($droplet->{name}) is now set up!");
   }
 
-  my $message = $exitcode == 0
-              ? "In-a-Box ($droplet->{name}) is now set up!"
-              : "Something went wrong setting up your box.";
+  my $message = "Something went wrong setting up your box.";
 
   if ($event->from_channel->isa('Synergy::Channel::Slack')) {
     $message .= " Here's the output from setup:";
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -36,6 +36,11 @@ has digitalocean_api_token => (
   required => 1,
 );
 
+has snippet_reactor_name => (
+  is       => 'ro',
+  isa      => 'Str',
+);
+
 has dobby => (
   is    => 'ro',
   lazy  => 1,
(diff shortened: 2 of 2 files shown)


lib/Synergy/Reactor/InABox.pm
```diff
@@ -202,6 +202,29 @@ async sub handle_status ($self, $event, $switches) {
   return await $event->reply("You don't seem to have any boxes.");
 }
 
+async sub _post_gitlab_snippet ($self, $payload) {
```
Davis-A, line 205: This could (should?) be moved to the Gitlab reactor and made public so other reactors can call into it.

lib/Synergy/Reactor/GitLab.pm
```diff
@@ -1109,6 +1109,27 @@ sub mr_report ($self, $who, $arg = {}) {
...
+    headers => {
+      'PRIVATE-TOKEN' => $self->api_token,
+    },
+
+    content_type => 'application/json',
+    content      => encode_json($payload),
+  );
+
+  unless ($res->is_success) {
+    die "error creating snippet on gitlab " . $res->as_string;
```
rjbs, line 1125:
```suggestion
    die "error creating snippet on GitLab: " . $res->as_string;
```

lib/Synergy/Reactor/GitLab.pm
```diff
@@ -1109,6 +1109,27 @@ sub mr_report ($self, $who, $arg = {}) {
...
+  }
+
+  my $json = $res->decoded_content(charset => 'undef');
```
rjbs, line 1128: undef should not be quoted, it is a keyword:

```suggestion
  my $json = $res->decoded_content(charset => undef);
```

It's used here to say that no charset decoding should be performed, only decompression.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -422,9 +422,21 @@ async sub _setup_droplet ($self, $event, $droplet, $key_file, $args = []) {
     return await $event->reply("In-a-Box ($droplet->{name}) is now set up!");
   }
 
-  my $message = $exitcode == 0
-              ? "In-a-Box ($droplet->{name}) is now set up!"
-              : "Something went wrong setting up your box.";
+  my $message = "Something went wrong setting up your box.";
+
+  # send error message as a snippet
+  if ($self->hub->reactor_named('gitlab')) {
```
rjbs, line 428: The reactor name to use should be a configuration option.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -422,9 +422,21 @@ async sub _setup_droplet ($self, $event, $droplet, $key_file, $args = []) {
...
+    my $username = $event->from_user->username;
+    my $payload;
+    $payload->{title} = "In-a-Box setup failure ($droplet->{name}) $username";
+    $payload->{file_name} = "In-a-Box-setup-failure-$droplet->{name}-$username.txt";
```
rjbs, line 432: Won't the username already be in the droplet name?  (I think so, but am not sure.)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -422,9 +422,21 @@ async sub _setup_droplet ($self, $event, $droplet, $key_file, $args = []) {
...
 
-  my $message = $exitcode == 0
-              ? "In-a-Box ($droplet->{name}) is now set up!"
-              : "Something went wrong setting up your box.";
+  my $message = "Something went wrong setting up your box.";
+
+  # send error message as a snippet
+  if ($self->hub->reactor_named('gitlab')) {
+    my $username = $event->from_user->username;
+    my $payload;
```
rjbs, line 430: It's a nit, but I would suggest `my %payload`.  Then you can drop a bunch of `->` operators.  This is more efficient, but not importantly so.  On the other hand, it is easier to read.  Then later, to pass a reference, you can just replace `$payload` with `\%payload`.

rjbs:
I've made some comments.  It took me a moment to realize that the behavior is:
* if GitLab, use that
* if not, and if current channel is Slack, use Slack snippet
* failing that, just use a normal reply

I think you should add a comment to that effect, and you can drop the comment that just says "post a GitLab snippet", since the code is clear.

rjbs (changes requested):

lib/Synergy/Reactor/GitLab.pm
Davis-A: I'll note that I cribbed that handling from Dobby/Client.pm, so that bug exists there in two places then: 

https://github.com/fastmail/Synergy/blob/main/lib/Dobby/Client.pm#L66
https://github.com/fastmail/Synergy/blob/main/lib/Dobby/Client.pm#L86

rjbs: WOW.  What on earth did I do there?  Thanks, I'll fix it.

rjbs:
I meant to add:  The commit messages do not match the usual style, in which almost all commits begin with the reactor or channel name _exactly_, meaning with matching case.

Davis-A:
Thanks for the extremely thorough review!  I think fixed everything requested

rjbs:
Oho!  So, your commit https://github.com/fastmail/Synergy/pull/212/commits/283c8478f5034b7dcdaf16aca2acec73adb17497 is a good fix, as it eliminates a weird never-happens case.  You're fixing my bug, but you're fixing it not the way I intended.  (How could you know??)

My intent was that the snippet would always be posted, so that you could always see what happened, just in case.  You can take 120s and think about whether that's a better idea.  I'm just not sure.

Anyway, this all looks good!  Thanks for the improvements.

rjbs (approved):

Davis-A:
> My intent was that the snippet would always be posted, so that you could always see what happened, just in case

I'm pretty happy only getting it if it fails.  For now i'll merge this, if you really want it to be all the time throw me a task and chuck the Fryday label on it and i'll do it soon.

PR #214, opened 2024-01-31
InABox: Grab 200 items per page for ssh keys

Davis-A:
We were generating a bunch of additional packer ephemeral keys and that pushed the fminabox key to the second page.  This is an easy bandaid until Dobby supports pagination


rjbs:
Thanks!  Rather than a stopgap, I went and implemented pagination in #215 since it was easy.  Thanks, Dobby!

PR #215, opened 2024-01-31
InABox: cope with paginated data for ssh_keys

rjbs:
This adds a Dobby method to get every page of results, concatenating the Array found in a given key from each page.  Then it uses it for SSH keys.

### DIFF
--- a/lib/Dobby/Client.pm
+++ b/lib/Dobby/Client.pm
@@ -142,20 +142,6 @@ async sub delete_url ($self, $path) {
   return;
 }
 
-async sub _get_snapshot ($self, $version) {
-  my $data = await $self->json_get('/snapshots?per_page=200');
-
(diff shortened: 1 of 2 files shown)


Davis-A (approved):

Davis-A:
Should probably also update this InABox method too :) 

```
async sub _get_snapshot ($self, $version) {
  my $dobby = $self->dobby;
  my $data  = await $dobby->json_get('/snapshots?per_page=200');

  my ($snapshot) = sort { $b->{name} cmp $a->{name}
                       || $b->{created_at} cmp $a->{created_at} }
                   grep { $_->{name} =~ m/^fminabox-\Q$version\E/ }
                   $data->{snapshots}->@*;

  if ($snapshot) {
    return $snapshot;
  }

  Synergy::X->throw_public("no snapshot found for fminabox-$version");
}
```

PR #216, opened 2024-02-05
InABox: make the DO ssh key name configurable

Davis-A:
The 'fminabox' key is defunct.  I've added synergy's key as 'synergy' so it's clear where that key is.  And this makes it configurable so if we ever have a future fminabox where we don't purge cloud-init then it can be configurable


marcbradshaw (approved):

rjbs:
This broke the test suite.  I have fixed it in 39ade23

PR #221, opened 2024-02-08
track whether events are completely handled

rjbs:
The main thrust here is that an Event now has a new attribute called "completeness", which is a Future that is ready when every PotentialReaction's returned future has become ready.  This can be used for a number of things, but most importantly determining how many events are still "in flight" on the event loop, which can be used for diagnostic reporting (like the changes added to the DiagnosticHandler) or for taking action based on the system's state.

The most important use now is that the Synergy::Tester system now waits until all events are complete, rather than all "todo" items in the input program have been injected.  This means we can eliminate arbitrary waits to avoid races.


wolfsage:
nit: Commit message for fb2f99df9aa:

  This means no more running a no-program testergize to get a result just
  to extract the hub to then run a program on after further setup.  It
  just makes sense.

  "run a program on after" -> "run a program after further setup" maybe

wolfsage:
nit: Commit 46cff8f842c7 (tests: move "use lib" into .yathrc...

  There's an unrelated fix in t/lib/Synergy/Tester.pm for whether @$todo has values or not

wolfsage:
cull events in flight timer -> safety? Feel like it should warn if it ever cleans stuff
  but I suppose in normal operations it might hit things before the future that'll clean them
  up does...

wolfsage:
So synergy has about 10 listeners that listen to every message
  This means that we're probably going to call _cull... 10 times each time doing a grep across
  the entire list and reassigning the list

  This is probably nbd but might cause a slight increase in processing times. I bet it will
  be insignificant but worth thinking about.

wolfsage:
nit: Commit 4ea724c2ed0e0ffb66551fca "Hub: small stylistic tweaks"

  This removes a comment but doesn't mention removing the comment or why it can now        
  be removed.

wolfsage:
nit: Still using 'result->synergy' in a bunch of tests that have a $synergy now after conversion
in the final commit

rjbs:
I have incorporated suggestions, thank you very much!

wolfsage:
> > So synergy has about 10 listeners that listen to every message This means that we're probably going to call _cull... 10 times each time doing a grep across the entire list and reassigning the list
> 
> I think you're mistaken here? The culler runs when the `Future->wait_all` across all the listeners completes, so only one for each _message_, not each _reaction_. Or do I misunderstand?

I'm totally mistaken. Once per message. Should be fine because most messages are people talking especially if we ignore bots before we even get to this point.

PR #222, opened 2024-02-09
InABox: Check snapshot regions before creating box

Davis-A:
When a snapshot is unavailable in the users datacentre it's much nicer to reply with a better response than "Something weird happened..."

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -253,6 +253,24 @@ async sub handle_create ($self, $event, $switches) {
   # the other does not, so here we'll just admit defeat and do them in
   # sequence. -- michael, 2020-04-02
   my $snapshot = await $self->_get_snapshot($version);
+  my %snapshot_regions = map {; $_ => 1 } $snapshot->{regions}->@*;
+
+  unless ($snapshot_regions{$region}) {
+    my $compatible_regions =
+      join ', ',
+      grep { $snapshot_regions{$_} } $self->box_datacentres->@*;
+
+    if ($compatible_regions) {
+      return await $event->reply(
+        "I'm unable to create snapshot in region '$region'.  Available compatible regions are $compatible_regions."
+      );
+    }
+
+    return await $event->reply(
+      "I'm unable to create snapshot in region '$region'.  Unfortunately this snapshot is not available in any of my configured regions"
+    );
+  }
+
   my $ssh_key  = await $self->_get_ssh_key;
 
   my $username = $user->username;
--- a/t/inabox.t
+++ b/t/inabox.t
@@ -215,6 +215,7 @@ my %CREATE_RESPONSES = (
     snapshots => [{
       id => 42,
       name => 'fminabox-bullseye-20200202',
+      regions => [ 'nyc3', 'sfo3'],
     }]
   }),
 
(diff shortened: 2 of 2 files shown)


Davis-A:
The functionality is there.  But i need to update the test suite before merging.

Davis-A:
Tests fixed so i think this is good to go

lib/Synergy/Reactor/InABox.pm
```diff
@@ -246,6 +246,27 @@ async sub handle_create ($self, $event, $switches) {
...
+    await $self->_get_snapshot_regions($snapshot->{id});
+
+  unless ($snapshot_regions{$region}) {
+    my $compatible_regions =
+      join ', ',
+      grep { $snapshot_regions{$_} } $self->box_datacentres->@*;
+
+    if ($compatible_regions) {
+      return await $event->reply(
+        "Unable to create snapshot in region '$region.  Available compatible regions are $compatible_regions."
```
wolfsage, line 261: nit: missing closing quote after `'$region`

Same below

wolfsage:
Nice, though you aren't testing for the new error conditions? Would be nice to!

lib/Synergy/Reactor/InABox.pm
rjbs: (yup)

lib/Synergy/Reactor/InABox.pm
```diff
@@ -246,6 +246,27 @@ async sub handle_create ($self, $event, $switches) {
   # the other does not, so here we'll just admit defeat and do them in
   # sequence. -- michael, 2020-04-02
   my $snapshot = await $self->_get_snapshot($version);
+
+  # check snapshot is available in the users region
+  my %snapshot_regions = map {; $_ => 1 }
+    await $self->_get_snapshot_regions($snapshot->{id});
```
rjbs, line 252: This seems off to me.  Above, with this line…

```perl
my $snapshot = await $self->_get_snapshot($version);
```

…it seems like we should have a snapshot already retrieved.  Then, `_get_snapshot_regions` goes and fetches it again.  Meanwhile, the API reference suggests that the objects returned from `_get_snapshot` should *already* have the region data.  That would mean that this branch is adding another round trip and another place for things to fail.

Can you, instead, look at `$snapshot->{regions}` without making another request?

rjbs (changes requested):
You're going to have to change this when you rebase after the big test improvements, but it should be easy.

But also, I thing you've got more API requests than are needed for this?  And testing the failure mode is probably a good idea, as Matthew said.

lib/Synergy/Reactor/InABox.pm
Davis-A: I am indeed an idiot.  For some reason my brain decided that `$snapshot` was just an id, but clearly it's a hash because i'm getting the id from it 🤦

rjbs: If this is the worst mistake you make all week, you'll be doing great. 😄

Davis-A:
Okie dokie.  Re-done without my 🤦 error.

Also added tests for the snapshot check too.

```
adavis@carbon:~/src/Synergy$ yath t/inabox.t

** Defaulting to the 'test' command **

( PASSED )  job  1    t/inabox.t

                                Yath Result Summary
-----------------------------------------------------------------------------------
     File Count: 1
Assertion Count: 33
      Wall Time: 3.28 seconds
       CPU Time: 3.69 seconds (usr: 0.12s | sys: 0.04s | cusr: 2.99s | csys: 0.54s)
      CPU Usage: 112%
    -->  Result: PASSED  <--
```

rjbs (approved):
Love it.

PR #226, opened 2024-03-07
InABox: lowercase before parsing switches

Davis-A:
It's common for me to use a phone to ask Synergy to spin up an Inabox, so by the time i'm at my desk it's ready to go.  My phone also likes to capitalise words.  And i'm a bit tired of stubbing my proverbial toe after `box create /version Bookworm` and being told there is no 'Bookworm' version available because of case sensitivity.

This seems like a reasonable local change to stop me bumping into this.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -154,6 +154,7 @@ command box => {
     return await $event->error_reply("usage: box [status|create|destroy|shutdown|poweroff|poweron|vpn]");
   }
 
+  $args = lc($args) if $args;
   my ($switches, $error) = parse_switches($args);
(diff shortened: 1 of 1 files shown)


wolfsage:
I don't think we should do this. This will lowercase args passed into every users' custom setup scripts (if they have them), and maybe some of those args take upper case characters!

Is it only about Bookworm/bookworm? We could lowercase version names when doing comparisons. If it's other things... we could come up with solutions that might not break all args

Davis-A:
> This will lowercase args passed into every users' custom setup scripts

That's a good call.  I'd completely forgotten that you can pass arguments to setup scripts.  I'll close this out in favour of another method that just affects the tag and version

PR #228, opened 2024-03-11
Dobby replicate images to regions

Davis-A:
I'm not sure what should be handled by what.  I can totally see an argument for Dobby having a method transfer_image_to_region()', note singular and have the InABox reactor handle transferring to multiple regions.

I can also see an argument to not change the box create code and instead have a timer that kicks off every 15 minutes to check if the image is in the right place and have it request the image transfer.

Basically I can think of a dozen ways so structure this and I'm not sure what the preferred option is.  

I can also see an argument for this case is pretty rare, and keep it as having a human press the button if this happens.


wolfsage:
Yeah I'm not sure... We know roughly when new droplets are created nightly, so we could have synergy wake up once a day 2 hours after that and see if it's in all the regions, and if not attempt to fix it. *And* we could also have this thing here which will catch things earlier, maybe...

I think putting it in Dobby and having an argument to handle error checking or not differently makes sense. You could wrap that with two methods:
* transfer_image_to_region_strict
* transfer_image_to_region_lax

Or just `transfer_image_to_region($image, $region, { allow_in_progress => 1 });`

Your / rik's call :)

PR #229, opened 2024-03-11
InABox: add /datacentre switch

Davis-A:
Just a convenience thing I should have created when added better replies when the region isn't available in fb7f3ca.

Also fixed the wording, which was apparently always wrong.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -161,7 +161,16 @@ command box => {
 
   # This should be simplified into a more generic "validate and normalize
   # switches" call. -- rjbs, 2023-10-20
-  for my $k (qw( version tag size )) {
+  # Normalize datacentre
+  if (exists $switches{datacentre} && exists $switches{datacenter}) {
+    return await $event->error_reply("You can't use /datacentre and /datacenter at the same time!");
+  }
+
+  if (exists $switches{datacenter}) {
+    $switches{datacentre} = delete $switches{datacenter};
+  }
+
+  for my $k (qw( version tag size datacentre )) {
     next unless $switches{$k};
     $switches{$k} = $switches{$k}[0];
   }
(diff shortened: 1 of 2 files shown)


lib/Synergy/Reactor/InABox.pm
```diff
@@ -264,12 +266,12 @@ async sub handle_create ($self, $event, $switches) {
 
     if ($compatible_regions) {
       return await $event->reply(
-        "I'm unable to create snapshot in region '$region'.  Available compatible regions are $compatible_regions."
+        "I'm unable to create an fminabox in region '$region'.  Available compatible regions are $compatible_regions.  You can use /datacentre switch and specify a compatible one"
```
wolfsage, line 269: nit: s/and specify/to specify/.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -161,7 +161,7 @@ command box => {
 
   # This should be simplified into a more generic "validate and normalize
   # switches" call. -- rjbs, 2023-10-20
-  for my $k (qw( version tag size )) {
+  for my $k (qw( version tag size datacentre datacenter )) {
```
wolfsage, line 164: This is okay, but what might be better is:
```
if ($switches{datacentre} && $switches{datacenter}) {
  # Reply with an error saying you can't specify both
}
$switches{datacentre} //= delete $switches{datacenter};

for my $k (qw( version tag size datacentre )) {
   ...
```
That way, later you just have:
```
my $region = $switches->{datacenter} // $self->_region_for_user($user);
```

But it's a nit really

wolfsage (approved):
Nice, thanks. Also thanks for adding the ~proper~ US spelling

lib/Synergy/Reactor/InABox.pm
wolfsage: I've force pushed a change for this

lib/Synergy/Reactor/InABox.pm
wolfsage: I've force pushed a change for this

wolfsage:
I've fixed up, merged, and deployed this. Thanks!

PR #230, opened 2024-03-11
HTTPServer: specify IP address

Davis-A:
Wasn't sure what ip address Synergy was listening on, then realised she was listening on everything.  Keep that default but allow specifying specific IPs


wolfsage:
Sure

wolfsage (approved):

PR #233, opened 2024-04-18
TimeClock: avoid blocking and dying (those are bad)

rjbs:
The old code has this:

    my $report = $report_reactor->begin_report($report{$which}, $user);

    my ($text, $alts) = $report->get;

    next unless defined $text; # !? -- rjbs, 2019-10-29

    $Logger->log([ "sending %s report for %s", $which, $user->username ]);
    $channel->send_message_to_user($user, $text, $alts);

That ->get means the whole `check_for_shift_changes` method, called by an every-15m timer, would block.  Worse, it would convert any failed future into an exception, and because `check_for_shift_changes` isn't called from a reaction, it isn't handled by the hub's usual exception handler.

This fix is fairly crude, but it should avoid blocking and crashing. Future work should...

* convert this stuff to async/await!
* add an uptime gauge to the Prometheus endpoint
* improve how we log errors especially

### DIFF
--- a/lib/Synergy/Reactor/TimeClock.pm
+++ b/lib/Synergy/Reactor/TimeClock.pm
@@ -432,16 +432,29 @@ sub check_for_shift_changes ($self) {
 
       next unless $will_send;
 
+      $Logger->log([ "kicking off %s report for %s", $which, $user->username ]);
       my $report = $report_reactor->begin_report($report{$which}, $user);
 
-      my ($text, $alts) = $report->get;
+      $report
+        ->else(sub (@error) {
+          $Logger->log([
+            "ERROR sending %s report to %s: %s",
+            $which,
+            $user->username,
+            "@error",
+          ]);
+          Future->done;
+        })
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/TimeClock.pm
```diff
@@ -432,16 +432,29 @@ sub check_for_shift_changes ($self) {
...
+          $Logger->log([
+            "ERROR sending %s report to %s: %s",
+            $which,
+            $user->username,
+            "@error",
+          ]);
+          Future->done;
+        })
+        ->then(sub ($text, $alts) {
+          return Future->done unless defined $text; # !? -- rjbs, 2019-10-29
```
wolfsage, line 449: is this worth loggering or...?

wolfsage (approved):
Seems right

lib/Synergy/Reactor/TimeClock.pm
rjbs: Yes, added.

PR #234, opened 2024-05-13
PagerDuty: Split incident text blocks

lerlacher-fm:
We are putting all the incident text into one slack post text block, which is too much text for one block when there's too many incidents.

Split it up into one block per incident instead.

This is untested because every time I want to test a synergy change I have to figure out how to run Synergy again from scratch.

### DIFF
--- a/lib/Synergy/Reactor/PagerDuty.pm
+++ b/lib/Synergy/Reactor/PagerDuty.pm
@@ -999,22 +999,7 @@ sub _active_incidents_summary ($self) {
         PL_N('incident', $count)
       );
 
-      my (@text, @slack);
-
-      for my $incident (@incidents) {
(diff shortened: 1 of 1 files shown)


rjbs:
This code doesn't even compile, I'm afraid.

I think maybe on new line 1027 you just need to replace `@blocks` with `@$blocks`, but I'm not 100% sure.

Hey, wait, I was going to send this back to you, but I realize that I can at least test that this kinda works, with that change.  Let's do it…

There was another error: you had `push @blocks, { ... { ... }; };, — the semicolon was inside an inner hashref, no good.

With those errors amended away, this appears to function.  I've also changed the commit message to say `PagerDuty` with a capital D. 😉

I'd like to make sure you find it easy to hack on Synergy and then to test things, so I'll try to be in touch on that front soon.

PR #235, opened 2024-05-13
Update sample config and readme for local development

lerlacher-fm:
Update the README to remove pitfalls, and update the sample config in eg:

* Remove the linear reactor since without an API key synergy will crash when loading
* Add eg/users.yaml (copied from test data) as without one, synergy will crash when loading

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -16,7 +16,7 @@ You need to set up a configuration file in YAML or JSON format.  Look at the
 example file to see what config options are expected, because they're not
 documented yet.  Run the program like this, more or less:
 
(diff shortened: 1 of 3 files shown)


lerlacher-fm:
This is removing a couple of obstacles to locally spinning up an instance of Synergy.

More sample config is required to make it easy to add in reactors such as Slack and PagerDuty. Not sure how exactly to do that in the most ergonomic way.

rjbs (approved):
I think I should import my local practices into the docs, because I think they'll be useful for people other than me.  But that's more work.

This seems like a pure improvement, and we should Just Merge It.  Thanks for writing it!

PR #236, opened 2024-06-27
InABox: Add support for /region alias for /datacentre

lerlacher-fm:
I always type "/region" on the first try because I've spent too much time in the cloud so region is the natural concept in my head rather than datacentre. So let's make the robot accomodate the human.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -16,7 +16,8 @@ use Process::Status;
 use Synergy::CommandPost;
 use Synergy::Logger '$Logger';
 use Synergy::Util qw(bool_from_text reformat_help);
-use String::Switches qw(parse_switches);
+use String::Switches qw(parse_switches canonicalize_names);
+use JSON::MaybeXS;
 use Future::Utils qw(repeat);
 use Text::Template;
 use Time::Duration qw(ago);
@@ -159,18 +160,12 @@ command box => {
   my ($switches, $error) = parse_switches($args);
   return await $event->error_reply("couldn't parse switches: $error") if $error;
 
+  canonicalize_names($switches, {
+    datacenter => 'datacentre',
+    region => 'datacentre'
+  });
+
   my %switches = map { my ($k, @rest) = @$_; $k => \@rest } @$switches;
-
-  # This should be simplified into a more generic "validate and normalize
-  # switches" call. -- rjbs, 2023-10-20
-  # Normalize datacentre
-  if (exists $switches{datacentre} && exists $switches{datacenter}) {
-    return await $event->error_reply("You can't use /datacentre and /datacenter at the same time!");
-  }
-
-  if (exists $switches{datacenter}) {
-    $switches{datacentre} = delete $switches{datacenter};
-  }
 
   for my $k (qw( version tag size datacentre )) {
     next unless $switches{$k};


lib/Synergy/Reactor/InABox.pm
```diff
@@ -159,15 +159,23 @@ command box => {
 
   my %switches = map { my ($k, @rest) = @$_; $k => \@rest } @$switches;
 
+  if (!exists $switches{datacentre}) {
```
wolfsage, line 162: I think the end goal here is that datacentre, datacenter, and region all mean the same thing, and providing more than 1 of them is a mistake. I think we can simplify this greatly with:
```
  my @got = grep { exists $switches{$_} } qw(datacenter datacentre region);

  if (@got > 1) {
    die "The following options are mutually exclusive: @got\n";
  }

  $switches{datacentre} = delete $switches{$got[0]};
```

Thoughts?

lerlacher-fm: Yes, that looks much cleaner! Let me test if this works in all cases.

lerlacher-fm: I have tested this and slightly adjusted it to make it all work nicely.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -159,17 +159,17 @@ command box => {
...
   my %switches = map { my ($k, @rest) = @$_; $k => \@rest } @$switches;
 
-  # This should be simplified into a more generic "validate and normalize
-  # switches" call. -- rjbs, 2023-10-20
-  # Normalize datacentre
-  if (exists $switches{datacentre} && exists $switches{datacenter}) {
-    return await $event->error_reply("You can't use /datacentre and /datacenter at the same time!");
-  }
+  # Special handling for aliased datacenter / datacentre / region switch
+  my @got = grep { exists $switches{$_} } qw(datacenter datacentre region);
```
rjbs, line 163: See that comment of mine you're deleting?  About a more generic switch normalizer?  We have one now!  Actually, weirdly, we had one then.  I'm not sure why I didn't use it.  Maybe because it doesn't validate, but it *does* normalize.

You should import `canonicalize_names` from String::Switches at the top of this file.  Then you can write, above the `my %switches = map …` line:

```perl
canonicalize_names($switches, {
  datacenter => 'datacentre',
  region => 'datacentre'
});
```

(As a side note, our policy is to use the American English spelling for code and variables, but changing that will take a bit more work here, so let's let it slide for now.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -159,17 +159,17 @@ command box => {
...
-  if (exists $switches{datacenter}) {
-    $switches{datacentre} = delete $switches{datacenter};
+  if (@got > 1) {
+    return await $event->error_reply("The following options are mutually exclusive: @got\n");
+  }
+  elsif (@got > 0) {
+    $switches{datacentre} = delete $switches{$got[0]};
   }
 
+
```
rjbs, line 172: bogus extra space

lib/Synergy/Reactor/InABox.pm
```diff
@@ -159,17 +159,17 @@ command box => {
...
+  # Special handling for aliased datacenter / datacentre / region switch
+  my @got = grep { exists $switches{$_} } qw(datacenter datacentre region);
 
-  if (exists $switches{datacenter}) {
-    $switches{datacentre} = delete $switches{datacenter};
+  if (@got > 1) {
+    return await $event->error_reply("The following options are mutually exclusive: @got\n");
+  }
+  elsif (@got > 0) {
+    $switches{datacentre} = delete $switches{$got[0]};
```
rjbs, line 169: This won't be needed if you use the canonicalizer.  (The last switch will just win, but if we want to fix that, we fix it in the `map` statement, not here.  That will fix it for every switch.)

rjbs (changes requested):

lerlacher-fm:
@rjbs I have applied your suggestions, rebased, and gave it a very rudimentary test-drive.

rjbs:
@lerlacher-fm Thanks for humoring me, change looks good (and did indeed drop some complexity imho). 👍🏽

rjbs (approved):

PR #237, opened 2024-07-04
GitLab: Fix post_gitlab_snippet headers

Davis-A:
This has been broken since 98df89e when it was changed to use `hub->http_post`

Other than `Content` and `Content_Type` k:v pairs, `hub->http_request()` presumes all other k:v pairs are headers.

By calling with with headers => { foo=> bar }, the actual arguments into http_client->do_request() were

headers => { headers => { foo => bar } }
instead of
headers => { foo => bar }


rjbs:
Oof, sorry about that breakage, @Davis-A .  Clearly you've righted my wrong. 👍🏽

rjbs (approved):

PR #238, opened 2024-07-29
InABox: Add preference to omit /force when destroying active boxes

Davis-A:


lib/Synergy/Reactor/InABox.pm
```diff
@@ -782,4 +788,11 @@ __PACKAGE__->add_preference(
   description => 'should box creation run setup by default?',
 );
 
+__PACKAGE__->add_preference(
+  name      => 'destroy-always-force',
+  validator => async sub ($self, $value, @) { return bool_from_text($value) },
+  default   => 0,
+  description => 'Omit /force on box destroy when droplet is powered on',
```
rjbs, line 795:
```suggestion
  description => 'When destroying an active box, always act as if /force was passed',
```

rjbs (approved):
I made one very small suggestion: change the description of the preference.

PR #240, opened 2024-09-06
Inabox: destroy: Strictly validate arguments

wolfsage:
This way, you don't accidentally destroy the wrong box if you do

    synergy: box destroy /label foo /force

... when you meant ...

    synergy: box destroy /tag foo /force

The former would estroy your base box image, the latter would destroy the specifically tagged image.


rjbs:
👍🏽

rjbs (approved):
Looks good, go for it.

PR #242, opened 2024-09-20
Slack: rewrite send_file using new Slack APIs

Davis-A:
files.upload is deprecated, use new method of files.getUploadURLExternal then files.completeUploadExternal

### DIFF
--- a/lib/Synergy/External/Slack.pm
+++ b/lib/Synergy/External/Slack.pm
@@ -13,6 +13,7 @@ use IO::Async::Timer::Periodic;
 use Net::Async::HTTP;
 use Net::Async::WebSocket::Client;
 use Data::Dumper::Concise;
+use Defined::KV;
 
 use Synergy::Logger '$Logger';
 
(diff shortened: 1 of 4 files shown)


Davis-A:
I think this is right.  We only have a single example of sending a txt vpn config, so i hacked in a temporary example to send a catpic and it worked fine. 

```
diff --git a/lib/Synergy/Reactor/InABox.pm b/lib/Synergy/Reactor/InABox.pm
index 9173229..bc42d0b 100644
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -641,7 +641,9 @@ async sub handle_vpn ($self, $event, $switches) {
     droplet_host => $self->_box_name_for($event->from_user, ($is_default_box ? () : $tag)),
   });

-  await $event->from_channel->send_file_to_user($event->from_user, 'fminabox.conf', $config);
+  use Path::Tiny;
+  my $catpic = path('tmp/catpic.jpg')->slurp_raw();
+  await $event->from_channel->send_file_to_user($event->from_user, 'catpic.jpg', $catpic);

   await $event->reply("I sent you a VPN config in a direct message. Download it and import it into your OpenVPN client.");
 }
```

Davis-A:
I later realised we had other places calling `slack.files` api manually, eg for the Inabox setup log on failure, which is probably what you wanted me to look at when you created the task.  

You probably didn't even know the slack channel and slack external modules had send file methods or you'd probably have used them when adding the log!

Anyway, i've updated all the places to use `Synergy::External::Slack->send_file()`

I also added in support for the file title, though we don't use it anywhere.  Based on what I saw when testing it, it replaces the filename in places slack displays that (including when downloading the file, it's weird).

lib/Synergy/External/Slack.pm
```diff
@@ -355,32 +356,63 @@ sub _send_rich_text ($self, $channel, $rich) {
...
-    my $res = decode_json($http_res->decoded_content(charset => undef));
-    $f->done({
-      type => 'slack',
-      transport_data => $res
-    });
+  my $json = decode_json($res->content);
+  die "Could get upload url to send file to slack: " . $json->{error}
+    unless $json->{ok};
+
+  $res = await $self->hub->http_client->POST(
```
rjbs, line 382: This is okay, but I prefer that we use new variables when doing a new request.  So `my $get_url_res` and then `my $post_content_res` etc.  Otherwise it can be hard to know the binding of a variable at any given part of the sub without remembering it all at once.  (This is why const is good and perl is bad? :) )

lib/Synergy/External/Slack.pm
```diff
@@ -355,32 +356,63 @@ sub _send_rich_text ($self, $channel, $rich) {
...
+    files      => [{
+      id => $json->{file_id},
+      defined_kv(title => $title),
+    }],
   });
 
-  return $f;
+  # The docs say application/x-www-form-urlencoded
+  # OR application/json but the former does not work
+  $res = await $self->hub->http_client->POST(
```
rjbs, line 402: (again as above)

rjbs (approved):
This is good, thanks!

I made one small request about variable reuse.

PR #245, opened 2024-11-06
Gitlab: add pipelines command

rjlov:
Shows a list of recently run pipelines and their success status, with a link to the results.

The idea is to easily see when run-perl-tests has stopped / started failing.

### DIFF
--- a/lib/Synergy/Reactor/GitLab.pm
+++ b/lib/Synergy/Reactor/GitLab.pm
@@ -19,6 +19,7 @@ use Future::AsyncAwait;
 use JSON::MaybeXS;
 use Lingua::EN::Inflect qw(PL_N);
 use List::Util qw(all uniq);
+use Slack::BlockKit::Sugar -all => { -prefix => 'bk_' };
 use String::Switches;
 use Synergy::CommandPost;
 use Synergy::Logger '$Logger';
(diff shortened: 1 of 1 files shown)


rjbs:
A PR from @rjlov !  So exciting.  Let's review.

lib/Synergy/Reactor/GitLab.pm
```diff
@@ -906,6 +915,72 @@ async sub _page_for_search_string ($self, $text, $event, $arg = undef) {
...
+        );
+      }
+    }
+    my $job_url = sprintf(
+      "%s/fastmail/%s/-/jobs/%s",
+      $self->url_base,
+      uri_escape($project_name),
+      uri_escape($job->{id})
+    );
+    $reply .= sprintf "\-26%s \%-8s \%-20s \%s \%s\n",
```
rjbs, line 959: This sprintf (and one below) are very odd.  The `%` are escaped, but I don't think for any real reason.  Also, the first format in this string seems wrong, with the `-26` before the `%` instead of between it and the `s`.

lib/Synergy/Reactor/GitLab.pm
```diff
@@ -906,6 +915,72 @@ async sub _page_for_search_string ($self, $text, $event, $arg = undef) {
...
+        $mr_url = sprintf(
+          "%s/fastmail/%s/-/merge_requests/%s",
+          $self->url_base,
+          uri_escape($project_name),
+          uri_escape($mr)
+        );
+      }
+    }
+    my $job_url = sprintf(
+      "%s/fastmail/%s/-/jobs/%s",
```
rjbs, line 954: Normally, we try to make these things (Synergy reactors) *kinda* Fastmail-agnostic, so you could use it elsewhere.  That would mean making `hm` and `fastmail` options, or config options.  For now, I'll apply as is, and we can do something later.

rjbs (approved):
I think the sprintf-ing at least needs to be fixed, but I'm going to just apply this and fix it myself, so I can desk-check this right now.

PR #246, opened 2024-11-15
InABox: add 'box image' command

Davis-A:
`box image` will print information on what image a box would be created with when invoking box create.  Mostly concerned with the age of the snapshot


rjbs:
I liked it.

PR #249, opened 2024-11-29
Reactor/Linear: Implement 'triage /all'

lerlacher-fm:
The 'triage' command by default shows only unassigned issues, in order to assist with triaging tickets created or moved from other teams.

However, tickets we (in PLAT) create for triage through `synergy ++` will have the creator assigned. This is fine but we want to be able to query synergy for these tickets, so we do that with `synergy triage plat /all`.

Ticket: PLAT-2422

### DIFF
--- a/lib/Synergy/Reactor/Linear.pm
+++ b/lib/Synergy/Reactor/Linear.pm
@@ -712,14 +712,31 @@ command triage => {
     *triage `[TEAM]`*: list unassigned issues in the Triage state
 
     This lists (the first page of) all unassigned issues in the Triage state in
-    Linear.  You can supply an argument, the name of a team, to see only issues
-    for that team.
+    Linear.
+    You can supply an argument, the name of a team, to see only issues for that
+    team.
+    You can supply a switch, /all, to also see assigned issues.
     EOH
-} => async sub ($self, $event, $team_name) {
+} => async sub ($self, $event, $text) {
   await $self->_with_linear_client($event, async sub ($linear) {
     my %extra_search;
 
-    if (length $team_name) {
+    # I've implemented half of a getopt parser here... sorry
+    my $opt_all = 0;
+    my @args;
+
+    my @words = split ' ', $text if length $text;
+
+    for my $arg (@words) {
+      if ($arg eq '/all') {
+        $opt_all = 1;
+      } else {
+        push @args, $arg;
+      }
+    }
+
+    if (length $args[0]) {
+      my $team_name = $args[0];
       my $team = await $linear->lookup_team($team_name);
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/Linear.pm
```diff
@@ -712,14 +712,31 @@ command triage => {
...
+} => async sub ($self, $event, $text) {
   await $self->_with_linear_client($event, async sub ($linear) {
     my %extra_search;
 
-    if (length $team_name) {
+    # I've implemented half of a getopt parser here... sorry
+    my $opt_all = 0;
+    my @args;
+
+    my @words = split ' ', $text if length $text;
```
rjbs, line 728: Due to WEIRD SHIT FROM BEYOND TIME AND SPACE, you basically never want to put postfix modifiers on `my` declarations.  (The behavior you get is now fairly stable, but it's best to avoid, anyway.)  Instead, I'd always write:

```suggestion
    my @words = length $text ? (split /\s+/, $text) : ();
```

lib/Synergy/Reactor/Linear.pm
```diff
@@ -712,14 +712,31 @@ command triage => {
...
+
+    for my $arg (@words) {
+      if ($arg eq '/all') {
+        $opt_all = 1;
+      } else {
+        push @args, $arg;
+      }
+    }
+
```
rjbs, line 737: Probably you want an error case if `@args>1`.

rjbs (changes requested):
Please do add that error handling.  Silently ignored-but-unknown arguments are a source of bitter confusion.

lib/Synergy/Reactor/Linear.pm
```diff
@@ -712,14 +712,35 @@ command triage => {
...
+      if ($arg eq '/all') {
+        $opt_all = 1;
+      } elsif ($arg =~ /^\//) {
+        die "Unrecognized switch: $arg";
+      } else {
+        push @args, $arg;
+      }
+    }
+
+    die "too many args - please only pass a single arg for the team name (I got: " . join(' ', @args) . ')' if scalar @args > 1;
```
rjbs, line 740: This won't do what you want.  You'll get "the reactor crashed."  Most code, reporting an error to the user, uses `reply_error`, but there's a simpler way these days:

```suggestion
    if (@args > 1) {
      Synergy::X->throw_public("That's too many args - please only pass a single arg for the team name.");
    }
```

(The use of `scalar` here was also redundant.)

rjbs (approved):
Approved, except you gotta apply my commit.

PR #250, opened 2024-12-02
Page: Don't cc if the target is the source of the page

Davis-A:
I paged myself to test if my do not disturb settings were working as intended and was surprised to see a CC to go it.


rjbs:
I'm not sure I'm keen on this.  Why introduce a special case?  Either way, it produces a permanent record for the recipient as intended.

PR #251, opened 2024-12-12
PagerDuty: Exclude blank user from oncall list

lerlacher-fm:


rjbs:
Thanks!  I added one more commit, moving the id to config, to keep magic constants out of the code. It's live now.

PR #253, opened 2024-12-19
InABox: sort snapshots exclusively by created_at time

Davis-A:
Had a minor bug where the wrong snapshot was spun up because it matched `fminabox-$version` but had additional stuff afterwards that sorted earlier

This isn't a fix the issue it's possible to create non main builds that collide in snapshot name, but at least they'll time out being the snapshot synergy grabs in <24 hours


rjbs:
I would like to better understand the circumstances here.  I'll catch up on Slack.

PR #254, opened 2024-12-27
Reactor/Page: Implement "page oncall /all"

lerlacher-fm:
"page oncall /all" will page everyone in the final oncall escalation, as a last resort emergency flare.

This requires redoing the argument parsing in the page command, and implementing a function in the PagerDuty module to determine all the users in the oncall escalation.

### DIFF
--- a/lib/Synergy/Reactor/Page.pm
+++ b/lib/Synergy/Reactor/Page.pm
@@ -50,13 +50,38 @@ command page => {
 reserved for emergencies or at least "nobody is replying in chat!"
 
 • *page `WHO`: `MESSAGE`*: send this message to that person's phone or pager
+
+Special On-Call paging:
+
+• *page oncall: `MESSAGE`*: send this message to the current on-shift oncall
+• *page /all oncall: `MESSAGE`*: send this message to all oncall engineers
 END
 } => async sub ($self, $event, $rest) {
-  my ($who, $what) = $event->text =~ m/^page\s+@?([a-z]+):?\s+(.*)/is;
+  #my ($who, $what) = $event->text =~ m/^page\s+@?([a-z]+):?\s+(.*)/is;
+  # I've implemented half of a getopt parser here... sorry
+  my $opt_all = 0;
+  my @args;
 
-  unless (length $who and length $what) {
+  my @words = length $event->text ? (split /\s+/, $event->text) : ();
+
+  shift @words; # get rid of the "page" command word
+
+  for my $arg (@words) {
+    if ($arg eq '/all') {
+      $opt_all = 1;
+    } elsif ($arg =~ /^\//) {
+      Synergy::X->throw_public("page: Unrecognized switch: $arg");
+    } else {
+      push @args, $arg;
+    }
+  }
+
+  unless (@args > 1) {
     return await $event->error_reply("usage: page USER: MESSAGE");
   }
+
+  my $who = $args[0] =~ s/:$//r;
+  my $what = join ' ', @args[1 .. $#args];
 
   my @to_page;
   if ($who eq 'oncall') {
@@ -67,7 +92,12 @@ END
       # eagerly enough.  That should probably be made into a lazily cached
       # attribute like we use for (for example) the Slack user list.  But we
       # can do that in the future. -- rjbs, 2023-10-18
-      my @oncall_ids = await $pd->_current_oncall_ids;
+      my @oncall_ids;
+      if ($opt_all) {
+        @oncall_ids = await $pd->_escalation_oncalls;
+      } else {
+        @oncall_ids = await $pd->_current_oncall_ids;
+      }
       @to_page = grep {; $_ } map {; $pd->username_from_pd($_) } @oncall_ids;
     } else {
       $Logger->log("Unable to find reactor 'pagerduty'") unless $pd;
--- a/lib/Synergy/Reactor/Page.pm
+++ b/lib/Synergy/Reactor/Page.pm
@@ -57,7 +57,6 @@ Special On-Call paging:
 • *page /all oncall: `MESSAGE`*: send this message to all oncall engineers
 END
 } => async sub ($self, $event, $rest) {
-  #my ($who, $what) = $event->text =~ m/^page\s+@?([a-z]+):?\s+(.*)/is;
   # I've implemented half of a getopt parser here... sorry
   my $opt_all = 0;
   my @args;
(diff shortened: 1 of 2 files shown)


lerlacher-fm:
Mock testing on inabox shows this:

```
HEAD is now at 5e9edf0e page all oncall
[fminabox lerlacher@box-fastmail-01(@lerlacher-bookworm) Synergy] $ perl bin/synergy -c local.json 
LOG | listening on ip:port 0.0.0.0:8118
⬮⬮ term-rw ⬮⬮ Console channel online; type /help for help
⬮⬮ term-wo ⬮⬮ Console channel online
⬮⬮ twilio ⬮⬮ Console channel online
page oncall foo
LOG | message event from term-rw/u:lerlacher: page oncall foo
LOG | paging lerlacher
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ I don't know how to page lerlacher, sorry.
╰─────────────────────────────────────────────────────────────────────────────╯
page oncall /all foo
LOG | message event from term-rw/u:lerlacher: page oncall /all foo
LOG | paging lerlacher
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ I don't know how to page lerlacher, sorry.
╰─────────────────────────────────────────────────────────────────────────────╯

```

I believe that indicates that it's working.

lib/Synergy/Reactor/Page.pm
```diff
@@ -50,14 +50,39 @@ command page => {
...
+
+  my @words = length $event->text ? (split /\s+/, $event->text) : ();
 
-  unless (length $who and length $what) {
+  shift @words; # get rid of the "page" command word
+
+  for my $arg (@words) {
+    if ($arg eq '/all') {
+      $opt_all = 1;
+    } elsif ($arg =~ /^\//) {
```
rjbs, line 72: Normally we'd write `$arg =~ m{^/}` to avoid having to backwhack the `/`, but this is fine.

lib/Synergy/Reactor/Page.pm
```diff
@@ -50,14 +50,39 @@ command page => {
...
+      push @args, $arg;
+    }
+  }
+
+  unless (@args > 1) {
     return await $event->error_reply("usage: page USER: MESSAGE");
   }
 
+  my $who = $args[0] =~ s/:$//r;
+  my $what = join ' ', @args[1 .. $#args];
```
rjbs, line 84: This is fine, but if you set $who to `shift(@args) =~ s/:$//r` then you could just join `@args` below.  Quite a few Perl programmers (but not me) *hate* seeing the `$#array` syntax.

lib/Synergy/Reactor/Page.pm
```diff
@@ -67,7 +92,12 @@ END
       # eagerly enough.  That should probably be made into a lazily cached
       # attribute like we use for (for example) the Slack user list.  But we
       # can do that in the future. -- rjbs, 2023-10-18
-      my @oncall_ids = await $pd->_current_oncall_ids;
+      my @oncall_ids;
+      if ($opt_all) {
+        @oncall_ids = await $pd->_escalation_oncalls;
+      } else {
+        @oncall_ids = await $pd->_current_oncall_ids;
+      }
```
rjbs, line 100: We would normally write this with a ternary:

```perl
my @oncall_ids = $opt_all
               ? await $pd->_escalation_oncalls
               : await $pd->_current_oncall_ids;
```

Also, seems like if they both return ids, both method names should end in `_ids`.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -737,6 +737,23 @@ sub _current_oncall_ids ($self) {
...
+      my $final_escalation_rule = $data->{escalation_policy}{escalation_rules}[-1];
+
+      my @oncalls = map {; $_->{id} }
+                    grep {; $_->{type} eq 'user_reference' }
+                    $final_escalation_rule->{targets}->@*;
+
+      return Future->done(@oncalls);
+      });
+}
+
```
rjbs, line 756: Surely you'd be happier if this was an async method?

```
async sub _escalation_oncalls ($self) {
  ...
  my $data = await $self->_pd_request(...);
  ...
  return @oncall_ids;
}
```

? :)

rjbs (approved):
I think this can be merged.  You've got a job to do, and this will help you do it.

I left some "how we'd usually write this" comments about style or construction.

The biggest thing, though, is that you have done a fairly quick and dirty switch parser, when String::Switches already exists for this, and was built for Synergy.  The InABox reactor uses it, or see Linear::Client's use, which is more thorough.

Still, this could be merged as is to keep things going.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: I don't know! I patterned this after the other oncall subs which aren't.

lib/Synergy/Reactor/Page.pm
lerlacher-fm: > Also, seems like if they both return ids, both method names should end in _ids.

I considered this and rejected it when I was originally writing it but I can no longer figure out a justification for it.

lib/Synergy/Reactor/Page.pm
lerlacher-fm: Is this acceptable?

```
  my $who = shift @args;
  $who = ($who =~ s/:$//r);
  my $what = join ' ', @args;
```

lerlacher-fm:
Fixed most of the issues, please let me know if you have any further feedback.

lib/Synergy/Reactor/Page.pm
rjbs: It's fine, but kind of weird.  That is:  the `/r` switch means "return a new value, do not edit in place".  But you're using it, effectively, to edit in place.  The more idiomatic way to do what you've done is:

```perl
my $who = shift @args;
$who =~ s/:$//;
my $what = join ' ', @args;
```

These are fine details, not problems.  I am providing this comment as "hey this is how we would normally do this" and not "do this or it is not acceptable"!

lib/Synergy/Reactor/PagerDuty.pm
rjbs: Yeah, I think PagerDuty is one of the reacors that could use some more love converting to async.

My rule of thumb:  you want to always want async/await, never futures.  I think I've hit some counterexamples, but it is VERY weird.

rjbs:
Lookin' good, my approval stands.

PR #263, opened 2025-04-17
Synergy::BoxManager: inabox management outside the hub

rjbs:
Synergy is great and all, but sometimes it's nice to manage boxes from outside of chatops.  Making that possible was the real reason behind ever writing Dobby, and this commit realizes it.  This abstracts box management (create, destroy, setup, power events) from the InABox reactor.

The BoxManager needs to be given callbacks for chatops things like sending replies.  In a console environment, these can be replaced with something like "print".

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -13,6 +13,7 @@ use Future::AsyncAwait;
 
 use Dobby::Client;
 use Process::Status;
+use Synergy::BoxManager;
 use Synergy::CommandPost;
 use Synergy::Logger '$Logger';
 use Synergy::Util qw(bool_from_text reformat_help);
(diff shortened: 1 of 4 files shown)


rjbs:
Hm.  Doing a bit of downstream work from this has shown that while it's *great*, it's not *perfect*.  

Yet!

A few more commits to come later.

rjbs:
@wolfsage I think this is back to reviewable.

wolfsage:
> The "snapshot not here but is in region X" message drops the "but is
    in region X" part.

This is a bummer, but I think 99% of times the "in region X" is a static region we (some of us) will know, so I can live with it.

lib/Synergy/Reactor/InABox.pm
```diff
@@ -107,6 +119,13 @@ has known_box_versions => (
   },
 );
 
+# This is not here so we can set it to zero and get rate limited or see bugs in
+# production.  It's here so we can make the tests run fast. -- rjbs, 2024-02-09
+has post_creation_delay => (
```
wolfsage, line 124: I believe this has moved to BoxManager so isn't needed in InABox.pm anymore?

lib/Synergy/Reactor/InABox.pm
```diff
@@ -209,7 +228,8 @@ sub _determine_version_and_tag ($self, $event, $switches) {
 }
 
 async sub handle_status ($self, $event, $switches) {
-  my $droplets = await $self->_get_droplets_for($event->from_user);
+  my $boxman   = $self->box_manager_for_event($event);
+  my $droplets = await $boxman->get_droplets_for($event->from_user->username);
```
wolfsage, line 232: There's a lot of:
```
  my $username = $event->from_user->username;
  my $boxman   = $self->box_manager_for_event($event);
  await $boxman->take_droplet_action($username, ...
```
I wonder if it would be beneficial to have box_manager_for_event() pull the username out and bind the username to itself so it's never passed as an argument

wolfsage:
A few minor nitpicks and a thought on a light refactor to move the username into the box manager. Take 'em or leave 'em!

wolfsage (approved):
(Have not tested after all, if you want me to, please ask)

lib/Synergy/Reactor/InABox.pm
rjbs: I think we want to be able to do cross-user queries.  This may get slicker in the future somehow…

PR #267, opened 2025-04-23
InABox: box obliterate, a temporary command

rjbs:
With the box domain changed, we "need" a way to destroy old boxes that have a different domain.  This addresses that problem.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -129,6 +129,9 @@ my %command_handler = (
   poweron  => \&handle_poweron,
   image    => \&handle_image,
   vpn      => \&handle_vpn,
+
+  # XXX Temporary
+  obliterate => \&handle_obliterate,
 );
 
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/InABox.pm
```diff
@@ -341,6 +344,34 @@ async sub handle_destroy ($self, $event, $switches) {
...
 
+async sub handle_obliterate ($self, $event) {
+  my $ident = $switches->{ident};
+
+  unless ($ident) {
+    Synergy::X->throw_public("You need to provide `/ident xyz-abc.box.fastmaildev.com`.");
+  }
+
+  unless ($ident =~ qr{\.box\.fastmaildev\.com\z}) {
+    Synergy::X->throw_public("You need to provide `/ident xyz-abc.box.fastmaildev.com`.");
```
wolfsage, line 355: nit: This error message should be different

wolfsage (approved):

PR #268, opened 2025-04-23
Reactor/InABox.pm: Add 'label' alias for 'tag' switch

lerlacher-fm:
I like 'label'.

### DIFF
--- a/lib/Synergy/Reactor/InABox.pm
+++ b/lib/Synergy/Reactor/InABox.pm
@@ -179,6 +179,7 @@ command box => {
     datacenter  => 'datacentre',
     region      => 'datacentre',
     tag         => 'ident',
+    label       => 'ident',
   });
 
   my %switches = map { my ($k, @rest) = @$_; $k => \@rest } @$switches;


lerlacher-fm:
I made this PR because I wrote "box create /label" and got an error back. My instinct will always be to use "region" instead of "datacenter", and "label" or "name" instead of "tag", and it'll be the opposite for other people. As long as there's no ambiguity, I think it's better for the bot to understand synonyms instead of not.

lerlacher-fm:
@rjbs any rebuttal? :-)

rjbs:
Yeah, I just feel pretty strongly about not having synonyms.  Three names for one thing means that two names are unnecessarily unavoidable.  It also creates confusion when you see something called "A" that you've always called "B", and wonder how they differ.  I think we will do well to avoid them.  The impact in this particular scenario is probably limited, but I don't think the value is high enough to vary from standard practice of "one thing, one name".

(I don't think this applies to "datacenter" and "datacentre".)

I think "label" is a good name here, since it matches up with the very specific thing it turns into: a DNS label.  Meanwhile, "ident" is used by our tooling to refer to the syslog ident, so that was a bad choice here.  I'll roll out synchronized updates to `box` and Synergy next week and send an email about the update.

PR #269, opened 2025-05-02
InABox: add a "username" preference

rjbs:
It's my preference (ha ha) that everyone's Synergy username matches their unix username for InABox-managed machines.  That ship has apparently sailed, at least a little.  Rather than cause trauma by renaming a lot of stuff, this adds a workaround.


wolfsage:
Seems right, I have not tested, I assume you have!

With this I reckon I can set my pref to 'rjbs' and make boxes for you just for fun, but that seems like an okay silliness to allow

wolfsage (approved):

PR #271, opened 2025-07-02
overhaul "give oncall" command

lerlacher-fm:
New features for `give oncall` command:

1. Be smarter about picking who to overrride
2. Automatically determine how long to override if possible

### DIFF
--- a/lib/Synergy/Reactor/PagerDuty.pm
+++ b/lib/Synergy/Reactor/PagerDuty.pm
@@ -386,8 +386,9 @@ responder 'give-oncall' => {
   targeted  => 1,
   skip_help => 1, # provided by "help give oncall"
   matcher => sub ($self, $text, $event) {
-    my ($who, $dur) = $text =~ /\Agive\s+oncall\s+to\s+(\S+)\s+for\s+(.*)\s*$/i;
-    return [ $who, $dur ] if $who;
+    my ($who, $for, $dur) = $text =~ /\Agive\s+oncall\s+to\s+(\S+)(\s+for\s+(.*))?\s*$/i;
+    return [ $who, $dur ] if $who && $for;
+    return [ $who, undef ] if $who;
     return;
   },
 } => async sub ($self, $event, $who, $dur) {
@@ -413,26 +414,59 @@ responder 'give-oncall' => {
     return await $event->error_reply("Hmm, $they doesn't seem to be a PagerDuty user.");
   }
 
-  my $seconds = eval { parse_duration($dur) };
-  unless ($seconds) {
-    return await $event->error_reply("Hmm, I can't figure out how long you mean by '$dur'.");
-  }
-
-  if ($seconds > 60 * 60 * 8) {
-    return await $event->error_reply("Sorry, I can only give oncall for 8 hours max.");
-  }
-
-  if ($seconds < 60 ) {
-    return await $event->error_reply("That's less than a minute; did you forget a unit?");
-  }
-
   my $oncalls = await $self->_relevant_oncalls;
+  my $from_oncall;
 
-  if (@$oncalls > 1) {
+  if (@$oncalls == 1) {
+    $from_oncall = $oncalls->[0];
+  } else {
+    for my $oncall (@$oncalls) {
+      if ($self->username_from_pd($oncall->{user}{id}) eq $event->from_user->{username}) {
+        $from_oncall = $oncall;
+        last;
+      }
+    }
+  }
+
+  if (!$from_oncall) {
     return await $event->error_reply(
-      "Sorry; there's more than one person oncall right now, so I can't help you!"
+      "Sorry; there's more than one person oncall right now and you're not one of them, so I can't help you!"
     );
   }
+
+
+  my $start    = DateTime->now->add(seconds => 15);
+  my $seconds;
+
+  if ($dur) {
+    $seconds = eval { parse_duration($dur) };
+    unless ($seconds) {
+      return await $event->error_reply("Hmm, I can't figure out how long you mean by '$dur'.");
+    }
+
+    if ($seconds > 60 * 60 * 8) {
+      return await $event->error_reply("Sorry, I can only give oncall for 8 hours max.");
+    }
+
+    if ($seconds < 60 ) {
+      return await $event->error_reply("That's less than a minute; did you forget a unit?");
+    }
+  } else {
+    if ($from_oncall->{end}) {
+      # FIXME: This is a weird epicycle, because we need to convert the seconds
+      # into the end time against when we submit it to the API.
+      # We only really need the seconds for the reply at the end, so this should
+      # be rejiggered
+      my $end = $ISO8601->parse_datetime($from_oncall->{end});
+      $seconds = $end->clone->subtract_datetime_absolute($start)->seconds;
+    } else {
+      return await $event->error_reply(
+        "There's no shift end for the shift you're trying to override so I cannot " .
+        "figure out from context how long to override for. Please specify duration."
+      );
+    }
+  }
+
 
   my $schedule = $oncalls->[0]{schedule};
   unless ($schedule) {
@@ -440,7 +474,6 @@ responder 'give-oncall' => {
   }
 
   my $sched_id = $schedule->{id};
-  my $start    = DateTime->now->add(seconds => 15);
   my $end      = $start->clone->add(seconds => $seconds);
 
   my $overrides = await  $self->_pd_request_for_user(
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -342,10 +342,25 @@ command oncall => {
     return await $event->reply_error(q{It's just "oncall".  Did you want "give oncall"?});
   }
 
-  my @ids = await $self->_current_oncall_ids;
+  my $reply = "";
+  my $oncall_tree = await $self->_oncall_tree;
 
-  my @users = uniq map {; $self->username_from_pd($_) // $_ } @ids;
-  return await $event->reply('current oncall: ' . join(', ', sort @users));
+  my @levels = sort keys %$oncall_tree;
```
rjbs, line 348: I hope this never matters, but:  This will sort in ASCII string order.  Once we have a 10th level, it'll put 10 between 1 and 2.  You should use…

```perl
my @levels = sort { $a <=> $b } keys %$oncall_tree;
```

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -342,10 +342,25 @@ command oncall => {
...
+  my $max_level = @levels[-1 + @levels];
```
rjbs, line 349: You can (and should) write:

```perl
my $max_level = $levels[-1];
```

Negative indexes work from the big end backward.  This is the normal idiom.  Also, `@array[$idx]` will *often* do what you want here, but not always.  An individual item indexed out of an array uses the `$` sigil.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -342,10 +342,25 @@ command oncall => {
...
+
+  for my $level (@levels) {
+    my @users = uniq map {; $self->username_from_pd($_->{user}{id}) // $_->{user}{id} } $oncall_tree->{$level}->@*;
+
+    my $extra = ($level == 1) ? " (primary oncall)" :
+                ($level == 2) ? " (primary escalation)" :
+                ($level == $max_level) ? "(final escalation)" : "";
```
rjbs, line 356: House style is:

```perl
    my $extra = $level == 1          ? " (primary oncall)"
              : $level == 2          ? " (primary escalation)"
              : $level == $max_level ? " (final escalation)"
              :                        "";
```

(I don't know why I was unable to provide this as an acceptable suggestion.)

Note that this isn't just reformatting:  I also added a missing leading space to the max-level alternative.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -714,43 +729,61 @@ sub _check_long_maint ($self) {
...
+      my @oncalls = grep {; !$should_ignore{$_->{user}{id}} } $data->{oncalls}->@*;
+
+      for my $oncall (@oncalls) {
+        my $level = $oncall->{escalation_level};
+
+        if (!$oncall_tree{$level}) {
+          $oncall_tree{$level} = [];
+        }
+
+        push @{$oncall_tree{$level}}, $oncall;
```
rjbs, line 800: You don't need to ~fix~ change this, but normally we avoid circumfix dereference and write:

```perl
push $oncall_tree{$level}->@*, $oncall;
```

…so that the line can be read (and written) strictly left-to-right.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -714,43 +729,61 @@ sub _check_long_maint ($self) {
...
-  $self->_relevant_oncalls->then(sub ($oncalls) {
-    my @ids = map  {; $_->{user}{id} } @$oncalls;
-    return Future->done(@ids);
-  });
+# gets all currently on call users and filters for the appropriate escalation level.
+# level 1 == primary oncall
+# level 2 == primary escalation
+# final level == all hands on deck
+async sub _relevant_oncalls ($self, $level = 1) {
+  my %should_ignore = map {; $_ => 1 } $self->suppressed_user_ids;
```
rjbs, line 813: This line appears unused.  Should it be deleted, or used?

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -714,43 +729,61 @@ sub _check_long_maint ($self) {
...
-  return $self->_pd_request(GET => '/escalation_policies/' . $policy_id)
-    ->then(sub ($data){
-      my $final_escalation_rule = $data->{escalation_policy}{escalation_rules}[-1];
+  my $oncall_tree = await $self->_oncall_tree;
 
-      my @oncalls = map {; $_->{id} }
-                    grep {; $_->{type} eq 'user_reference' }
-                    $final_escalation_rule->{targets}->@*;
+  my @levels = sort keys %$oncall_tree;
+  my $max_level = @levels[-1 + @levels];
```
rjbs, line 832: Same comments apply as above:  should sort with `<=>` and use `[-1]` index.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -394,6 +394,13 @@ responder 'give-oncall' => {
   # of error checking up front here.
   $event->mark_handled;
 
+  if ($who eq 'escalation') {
+    my $escalations = await $self->_relevant_oncalls(2);
+
+    $who = $self->username_from_pd($escalations->[0]{user}{id});
+    $who ||= $escalations->[0]{user}{id}; # fallback mainly for debugging if we don't have a full user directory
+  }
```
rjbs, line 408: Nice, I liked this.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -385,8 +385,9 @@ responder 'give-oncall' => {
   targeted  => 1,
   skip_help => 1, # provided by "help give oncall"
   matcher => sub ($self, $text, $event) {
-    my ($who, $dur) = $text =~ /\Agive\s+oncall\s+to\s+(\S+)\s+for\s+(.*)\s*$/i;
-    return [ $who, $dur ] if $who;
+    my ($who, $for, $dur) = $text =~ /\Agive\s+oncall\s+to\s+(\S+)(\s+for\s+(.*)\s*)?$/i;
```
rjbs, line 388: I have no strong opinion on this, but prior to this commit, trailing space was ignored.  Now it's ignored only in the `for 1h` form.  You might want to move the final `)?` to the left of the `\s*`.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -426,20 +427,66 @@ responder 'give-oncall' => {
   }
 
   my $oncalls = await $self->_relevant_oncalls;
+  my $from_oncall;
 
-  if (@$oncalls > 1) {
+  if (@$oncalls == 1) {
+    $from_oncall = $oncalls->[0];
+  }
+  else {
```
rjbs, line 430: House style would cuddle the else (i.e., drop the newline between `}` and `else {`

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -426,20 +427,66 @@ responder 'give-oncall' => {
...
+      return await $event->error_reply("That's less than a minute; did you forget a unit?");
+    }
+  } else {
+    if ($from_oncall->{end}) {
+      # FIXME: This is a weird epicycle, because we need to convert the seconds
+      # into the end time against when we submit it to the API.
+      # We only really need the seconds for the reply at the end, so this should
+      # be rejiggered
+      my $end = $ISO8601->parse_datetime($from_oncall->{end});
+      $seconds = $end->clone->subtract_datetime_absolute($start)->seconds;
```
rjbs, line 469: Weird.  I have no obviously much simpler/better thing to do here, so… carry on! 🫡

rjbs (changes requested):
Mostly I made comments on style or idiom, but I think you *should* change:

1. most importantly, `should_ignore` being unused
2. sort of levels (twice)
3. subscripting for end of array

Other stuff, I would like to see changed, but wouldn't refuse to accept. ;)

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: I am pretty sure this can be fixed, but would require redoing the whole logic of the time calculations. Probably worth doing but not today.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: Well that's awkward because this diff is super messy. I will see if it can be made better.

rjbs:
Tests on devel fixed on main.  Wild, the `perldocker/perl-tester:devel` docker container has no /usr/bin/uptime even though the other versions of perl-tester do.

PR #275, opened 2025-07-21
Show incident notes

lerlacher-fm:
The 'incident' command now shows notes attached to live incidents.

This allows responders to leave notes for others who may need to check / respond to long-running incidents, and encourages them to do so.

### DIFF
--- a/lib/Synergy/Reactor/PagerDuty.pm
+++ b/lib/Synergy/Reactor/PagerDuty.pm
@@ -1120,47 +1120,45 @@ sub _announce_oncall_change ($self, $before, $after) {
   })->retain;
 }
 
-sub _active_incidents_summary ($self) {
-  return $self->_get_incidents(qw(triggered acknowledged))
-    ->then(sub (@incidents) {
-      return Future->done() unless @incidents;
+async sub _active_incidents_summary ($self) {
+  my @incidents = await $self->_get_incidents(qw(triggered acknowledged));
+  return unless @incidents;
 
-      my $count = @incidents;
-      my $title = sprintf("🚨 There %s %d %s on the board: 🚨",
-        PL_V('is', $count),
-        $count,
-        PL_N('incident', $count)
-      );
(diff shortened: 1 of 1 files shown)


lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -898,6 +898,13 @@ sub _get_incidents ($self, @statuses) {
   return Future->done(@results);
 }
 
+async sub _get_incident_notes($self, $incident_id) {
```
rjbs, line 901: house style puts a space between name and `(` of sig

rjbs: also, bummer this can't be specified in the `include[]` of extra data to provide!

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -1120,47 +1127,59 @@ sub _announce_oncall_change ($self, $before, $after) {
...
 
-      my $slack = {
-        blocks => bk_blocks(
-          bk_richblock(
-            bk_richsection(bk_bold($title)),
-            bk_ulist(@bk_items),
-          )
-        )->as_struct,
-      };
+    my @notes_blocks = ();
```
rjbs, line 1155: the initialization `= ()` isn't needed, and is usually best omitted, imo

rjbs (approved):
two tiny tiny nits

Thanks, it's very cheering to see somebody other than me rip out some of the sequencing code in favor of async/await!

I suspect that in the future, we may want to redo some of the @notes_blocks to use more block-list-form listing, but we can do that when we decide it's worth it! 

👍🏽

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: > also, bummer this can't be specified in the include[] of extra data to provide!

Indeed.

PR #276, opened 2025-08-11
allow report generators to use BlockKit

rjbs:


lib/Synergy/Channel/Console.pm
```diff
@@ -384,7 +384,7 @@ EOH
 
     my $content
       = $format eq 'text' ? $channel->_format_message_chonky(\%new_message)
-      : $format eq 'alts' ? YAML::XS::Dump($new_message{alts})
+      : $format eq 'alts' ? JSON::MaybeXS->new->convert_blessed->canonical->encode($new_message{alts})
```
wolfsage, line 387: This module doesn't load JSON::MaybeXS - it probably should

PR #278, opened 2025-08-14
Update dependencies

Davis-A:


rjbs (changes requested):
Presumably needs changes, if only to remove the merge commit.

PR #279, opened 2025-08-14
Inabox fix error message no box destroy

Davis-A:


rjbs:
OK to merge if merge commit is removed.

PR #287, opened 2026-03-20
Users table add id column

Davis-A:


rjbs:
Sorry, I'm keen to get to this, but it's been slightly below the fold for a few days.  It's on my list, though.

rjbs:
Meantime, how about you unbreak the build, so we have CI results?

```
couldn't determine document name for bin/migrate-user-ids
Add something like this to bin/migrate-user-ids:
# PODNAME: bobby_tables.pl at /usr/share/perl5/Pod/Weaver.pm line 143.
```

dzil won't build the dist without that fixed, so the tests can't run.

rjbs:
I've pushed a fix for the PODNAME issue.

lib/Synergy/UserDirectory.pm
```diff
@@ -129,24 +153,27 @@ sub load_users_from_database ($self) {
...
     });
+    $users{$username} = $user;
   }
 
-  my $identity_sth = $dbh->prepare('SELECT * FROM user_identities');
+  my $identity_sth = $dbh->prepare(
+    'SELECT u.username, ui.identity_name, ui.identity_value
+     FROM user_identities ui
+     JOIN users u ON ui.user_id = u.id'
+  );
```
rjbs, line 176: I find this hunk of changes weird.  The use of `defined_kv` above means that we will happily load users from the db who have no `id` field -- which would make sense if we did that to migrate, or if things fell back.

But instead, we then go to find their user identities by looking only identities that have non-null user ids, so any user who has no user id will have no identity.

I feel like the upgrade story here isn't quite coherent.  Possibly I'm confused, but I don't *think* so.  (Let's talk.)  The migration program does the db schema upgrade.  You won't have an `id` column until it's run.  If that's the case, this code would crash at runtime: we'd be querying for a does-not-exist column.

If migration *has* run, then `id` exists *and is populated*.  In that case, the `defined_kv` call doesn't make sense.

lib/Synergy/UserDirectory.pm
```diff
@@ -201,8 +234,14 @@ sub register_user ($self, $user) {
     q{VALUES (?,?,?,?)}
   ));
 
+  state $user_insert_with_id_sth = $dbh->prepare(join(q{ },
```
rjbs, line 237: Technically safe exactly as is, but using a `state` sth with `my` dbh invites future trouble if the dbh were able to change once in a while.  You'd have an sth pointing to a stale dbh.  Better to just make this and other statement handles `my`.  

I see now that you didn't make these `state`, but maybe while you're in here… ;-)

lib/Synergy/UserDirectory.pm
```diff
@@ -212,15 +251,29 @@ sub register_user ($self, $user) {
...
+    } else {
+      $user_insert_sth->execute(
+        $user->username,
+        $user->is_master,
+        $user->is_virtual,
+        $user->is_deleted,
+      );
+      $user_id = $dbh->last_insert_id;
+      $user->_set_id($user_id);
+    }
```
rjbs, line 273: In what case would we have a user with no id?

lib/Synergy/Role/HasPreferences.pm
```diff
@@ -186,7 +205,17 @@ role {
...
+      my %username_keyed;
+      for my $user_id (keys %$prefs) {
+        my $user = $ud->user_by_id($user_id);
+        unless ($user) {
+          $Logger->log(["HasPreferences: skipping prefs for unknown user id %s during load", $user_id]);
+          next;
+        }
+        $username_keyed{ $user->username } = $prefs->{$user_id};
+      }
+      $self->_load_preferences(\%username_keyed);
```
rjbs, line 218: Really not clear to me what this code (and the similar code above) is for.  Is this meant to live on forever?  Gets back to "do we have a world where users *might* have ids, but not *must* have ids?"

Feels like we should be planning for a world where every user always has an id, which eliminates… all? of this code?  We get there by taking down Synergy, running the upgrade, then bringing up the new version.

rjbs (changes requested):
Right now, this code feels like it's going to add complexity to a bunch of places that will last forever, instead of being a clean cutover with minimal leftovers.  I think we can accomplish this with left retained code.

PR #43, opened 2017-07-19
Add help for 'resume' command

wolfsage:

PR #291, opened 2026-08-12
Alert ergonomics

lerlacher-fm:
Two changes to PagerDuty reactor commands:

* the only thing you can ack is 'all': There is no point in enforcing 'ack all'. Any ack shall ack.
* snooze all: If 'all' is passed instead of an incident number, snooze everything

### DIFF
--- a/lib/Synergy/Reactor/PagerDuty.pm
+++ b/lib/Synergy/Reactor/PagerDuty.pm
@@ -619,24 +619,25 @@ command snooze => {
 
     my $id = $item->{id};
 
-    my $res = await $self->_pd_request_for_user(
+    my $res = eval { await $self->_pd_request_for_user(
       $event->from_user,
       POST => "/incidents/$id/snooze",
       { duration => $seconds }
-    );
+    ); };
+    my $error = $@;
 
     if (my $incident = $res->{incident}) {
       my $title = $incident->{title};
       push @snoozed, "#$id ($title)";
     } else {
-      push @errors, $res->{message};
+      push @errors, $error->message;
     }
   }
 
   my $reply = sprintf("Snoozed incidents for %s: \n%s", duration($seconds), join("\n", @snoozed));
 
   if (@errors) {
-    my $reply .= sprintf("\n\nUnfortunately we also received errors:\n%s", join("\n", @errors));
+    $reply .= sprintf("\n\nUnfortunately we also received errors:\n%s", join("\n", @errors));
   }
 
   return await $event->reply($reply);
@@ -699,7 +700,7 @@ sub _pd_request ($self, $method, $endpoint, $data = undef, $token = undef) {
     unless ($res->is_success) {
       my $code = $res->code;
       $Logger->log([ "error talking to PagerDuty: %s", $res->as_string ]);
-      return Future->fail('http', { http_res => $res });
+      return Future->fail($res->as_string, 'http', { http_res => $res });
     }
 
     my $data = decode_json($res->content);
--- a/lib/Synergy/Reactor/PagerDuty.pm
+++ b/lib/Synergy/Reactor/PagerDuty.pm
@@ -584,11 +584,14 @@ command resolve => {
 };
 
 command snooze => {
-  help => 'Snooze a single PagerDuty incident. Usage: snooze ALERT-NUMBER for DURATION',
-} => async sub ($self, $event, $rest) {
-  my ($num, $dur) = $rest =~ /^#?(\d+)\s+for\s+(.*)/i;
+  help => 'Snooze a PagerDuty incidents. Usage:
 
-  unless ($num && $dur) {
+  snooze ALERT-NUMBER for DURATION
+  snooze all for DURATION',
+} => async sub ($self, $event, $rest) {
+  my ($incident, $dur) = $rest =~ /^#?(\S+)\s+for\s+(.*)/i;
+
+  unless ($incident && $dur) {
     return await $event->error_reply(
       "Sorry, I don't understand. Say 'snooze INCIDENT-NUM for DURATION'."
     );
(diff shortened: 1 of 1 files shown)


lerlacher-fm:
@rjbs Before I spend time testing the implementation, do you have any objections to the behavior change I'm proposing here?

rjbs:
Nope!

lerlacher-fm:
Testing seems ok:

```
snooze #41478 for 1 hour
LOG | message event from term-rw/u:lerlacher: snooze #41478 for 1 hour
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Snoozed incidents for 1 hour: 
│ #Q0X18OP6VRORZF (phl-root-02 - movemate is restarting too fast)
╰─────────────────────────────────────────────────────────────────────────────╯
snooze all for 48 hours
LOG | message event from term-rw/u:lerlacher: snooze all for 48 hours
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Snoozed incidents for 2 days: 
│ #Q0X18OP6VRORZF (phl-root-02 - movemate is restarting too fast)
╰─────────────────────────────────────────────────────────────────────────────╯

```

```
ack
LOG | message event from term-rw/u:lerlacher: ack
LOG | PagerDuty: acking incidents: {{[]}}
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Successfully acked 0 incidents. Good luck!
╰─────────────────────────────────────────────────────────────────────────────╯
ack all
LOG | message event from term-rw/u:lerlacher: ack all
LOG | PagerDuty: acking incidents: {{[]}}
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Successfully acked 0 incidents. Good luck!
╰─────────────────────────────────────────────────────────────────────────────╯
ack nothing
LOG | message event from term-rw/u:lerlacher: ack nothing
LOG | PagerDuty: acking incidents: {{[]}}
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Successfully acked 0 incidents. Good luck!
╰─────────────────────────────────────────────────────────────────────────────╯
```

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -602,32 +601,41 @@ command snooze => {
...
+      $event->from_user,
+      POST => "/incidents/$id/snooze",
+      { duration => $seconds }
     );
+
+    if (my $incident = $res->{incident}) {
+      my $title = $incident->{title};
+      push @snoozed, "#$id ($title)";
+    } else {
+      push @errors, $res->{message};
```
themarcuslove: _pd_request fails the future on non-2xx, so it never returns a hash with message — the else at 628 is unreachable, and the bare await propagates the failure out of the command. snooze all for 1h where #42 got resolved by someone else mid-batch: #41 stays snoozed in PD, #43 is never attempted, and the user just gets "My PagerDuty reactor crashed (in the background)".

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -602,32 +601,41 @@ command snooze => {
...
   }
 
-  my $msg = $res->{message} // 'nothing useful';
+  my $reply = sprintf("Snoozed incidents for %s: \n%s", duration($seconds), join("\n", @snoozed));
 
-  return await $event->reply(
-    "Something went wrong talking to PagerDuty; they said: $msg"
-  );
+  if (@errors) {
+    my $reply .= sprintf("\n\nUnfortunately we also received errors:\n%s", join("\n", @errors));
```
themarcuslove: my $reply at 635 declares a fresh variable inside the if, so the error text goes to a throwaway and never reaches the user.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -522,12 +522,8 @@ responder 'give-oncall' => {
 };
 
 command ack => {
-  help => '*ack all*: acknowledge all triggered alerts in PagerDuty',
+  help => '*ack*: acknowledge all triggered alerts in PagerDuty',
 } => async sub ($self, $event, $rest) {
```
themarcuslove, line 526: $rest is bound here and never read again, so ack 42 acks the whole board silently — same for ack mine, ack banana. Bare ack meaning "all" is right, but acking all when trying to ack one is probably bad.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -602,32 +601,41 @@ command snooze => {
 
   my @incidents = await $self->_get_incidents(qw(triggered acknowledged));
 
-  my ($relevant) = grep {; $_->{incident_number} == $num } @incidents;
-  unless ($relevant) {
-    return await $event->error_reply("I couldn't find an active incident for #$num");
+  # select a single incident if we
+  my @relevant = ($incident =~ /\d+/) ? grep {; $_->{incident_number} == $incident } @incidents : @incidents;
```
themarcuslove, line 609: /\d+/ isn't anchored, so it's asking "is there a digit anywhere" rather than "is this an incident number". mine has no digit, falls through to the all-branch.

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -584,11 +580,14 @@ command resolve => {
...
 
 command snooze => {
-  help => 'Snooze a single PagerDuty incident. Usage: snooze ALERT-NUMBER for DURATION',
+  help => 'Snooze a PagerDuty incidents. Usage:
+
+  snooze ALERT-NUMBER for DURATION
+  snooze all for DURATION',
 } => async sub ($self, $event, $rest) {
-  my ($num, $dur) = $rest =~ /^#?(\d+)\s+for\s+(.*)/i;
+  my ($incident, $dur) = $rest =~ /^#?(\S+)\s+for\s+(.*)/i;
```
themarcuslove, line 592: A bare snooze gets $rest = undef from CommandPost, so this matches a regex against undef and probably logs a warning.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: Good catch - I can reproduce this by forcing a `Future->fail` in `_pd_request`. I've added a fix commit which isn't elegant (catching failed futures ergonomically seems to be a missing feature in Future::AsyncAwait) but should work.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: Good catch, I'm surprised compile testing didn't catch that.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: As the docs and the commits explain, we only have 'ack all' and nothing else.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: We don't care about 'mine' for acking.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: The warning does pop, but I'm not concerned about it - the behavior is fine

```
snooze all
LOG | message event from term-rw/u:lerlacher: snooze all
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Sorry, I don't understand. Say 'snooze INCIDENT-NUM for DURATION'.
╰─────────────────────────────────────────────────────────────────────────────╯
snooze all for
LOG | message event from term-rw/u:lerlacher: snooze all for
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Sorry, I don't understand. Say 'snooze INCIDENT-NUM for DURATION'.
╰─────────────────────────────────────────────────────────────────────────────╯
snooze
LOG | message event from term-rw/u:lerlacher: snooze
Use of uninitialized value $rest in pattern match (m//) at lib/Synergy/Reactor/PagerDuty.pm line 588.
╭─────┤ term-rw!sysop ├───────────────────────────────────────────────────────╮
│ Sorry, I don't understand. Say 'snooze INCIDENT-NUM for DURATION'.
╰─────────────────────────────────────────────────────────────────────────────╯
```

lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -602,32 +601,42 @@ command snooze => {
...
+      POST => "/incidents/$id/snooze",
+      { duration => $seconds }
+    ); };
+    my $error = $@;
+
+    if (my $incident = $res->{incident}) {
+      my $title = $incident->{title};
+      push @snoozed, "#$id ($title)";
+    } else {
+      push @errors, $error->message . ": " . $error->details->{http_res}->{message};
```
themarcuslove: The loop containment works — I tested eval { await ... } against Future::AsyncAwait 0.66 and it does run all incidents now, so the abandoned-batch problem is gone.

Two things wrong with this line though. ->{message} is a hash access on an HTTP::Response, which keeps the reason phrase in _msg — it's undef, so every error line ends in ": " with nothing after it and warns "Use of uninitialized value in concatenation".

And $error->message is now $res->as_string after the change at 699, so it's the whole response. One entry in @errors comes out as:

404 Not Found
Content-Type: application/json

{"error":{"message":"Incident Already Resolved","code":2001}}
:

lerlacher-fm: >  ->{message} is a hash access on an HTTP::Response, which keeps the reason phrase in _msg — it's undef

The right diagnosis here is that it should be `->message` to call the accessor function instead of `->{message}`.

It's superfluous anyway because `$error->message` already has everything we need.

lib/Synergy/Reactor/PagerDuty.pm
lerlacher-fm: I've now made it so only bare ack and "ack all" are accepted.

PR #294, opened 2026-09-02
convert the final reactors to CommandPost

rjbs:


lib/Synergy/Reactor/Vestaboard.pm
```diff
@@ -964,15 +825,13 @@ sub _post_payload ($self, $payload) {
 sub _pay_to_post_payload ($self, $event, $user, $payload) {
   my $lock = $self->current_lock;
   if ($lock && $lock->{locked_by} ne $user->username) {
-    $event->error_reply("Sorry, the board can't be changed right now!");
-    return;
+    return $event->error_reply("Sorry, the board can't be changed right now!");
```
Davis-A: I think this should be awaited.  Surprised me that it's not an async sub but if you follow it down you end up at 
```
  my $future = $send_cb->($text, $alts);

  $self->from_channel->note_reply($self, $future, $args);
  return $future;
```

ditto all the others

Davis-A (approved):
Approved with please fix those awaits on error replies

lib/Synergy/Reactor/Vestaboard.pm
rjbs: This doesn't need to be awaited.  `_pay_to_post_payload` is not an `async sub`, so it returns a future immediately so that the called can await it or otherwise sequence it.  If you look at its callers, _they_ await it, as is right and proper.

ditto all the others ;-)

That said, my response is basically "My code is correct, and you should understand both ways of doing this at all times."  A better response might be, "My code is correct, but I should stick to doing these things in a single way."

So I will apply that "reformatting" which will not fix a bug, and ship thusly.

PR #296, opened 2026-09-06
Tester: install the test support modules

Davis-A:
Synergy::Tester and the Synergy::Test:: modules only existed under t/lib, so a package-installed Synergy had nothing to test modules vendored elsewhere with. Move them into lib/ alongside Channel::Test and Logger::Test. Their prereqs are suggests rather than requires: you need them to use the test support, not to run a hub.

Test::CommandPost also called Test::Deep and Sub::Install by full name without ever loading them, which only worked because its one caller loaded Test::Deep itself.


dist.ini
```diff
@@ -48,6 +55,11 @@ Linear::Client = 0
 Dobby::BoxManager = 0
 Dobby::Client = 0
 
+IO::Async::Test = 0
+Net::EmptyPort = 0
+Sub::Install = 0
+Test::Deep = 0
```
rjbs, line 62: These shouldn't be Suggests, but TestRequires

rjbs (approved):
lgtm but you should change the one prereq category

PR #297, opened 2026-09-11
async/await: use it more, when we had been using manual sequencing

rjbs:


lib/Synergy/Reactor/PagerDuty.pm
```diff
@@ -1178,22 +1190,22 @@ async sub _active_incidents_summary ($self) {
...
     Accept        => 'application/vnd.pagerduty+json;version=2',
-  )->then(sub ($res) {
-    my $rc = $res->code;
+  );
+
+  my $rc = $res->code;
 
-    return Future->fail('That token seems invalid.')
-      if $rc == 401;
+  return await Future->fail('That token seems invalid.')
```
Davis-A: Bit strange but I'll allow it.  I'd probably have thrown an exception instead.  Or change the signature to return a tuple of `($value, $err)`

Davis-A (approved):

PR #299, opened 2026-09-18
Ask the channel, not Slack

Davis-A:
In the Yelling reactor we have a hook that checks the channel name, this is Slack specific.  This makes it difficult to test and makes the reactor incompatible with other channels. 

This PR adds a generic way to ask a channel the conversation name from a given `$event`.  And extends Channel::Test to set a channel name so we can actually have tests gated on names.  

While there also added an option to set is_public in Channel::Test so we can test reactors that have behaviour gated behind ` if ($event->is_public)` can also be unit tested.


rjbs:
Nit: commit typo, "Sub::Overrride" -- three r instead of two

Nit: commit typo, `didnt'`

These are my only complaints. 👍🏽

rjbs (approved):

